PR #2, opened 2021-06-17
Add RFC0004 - Deferred block syntax

leonerd:


nwc10:
I've run out of time today to deal with this (and the other admin implications of having 3 actual RFCs here!) but summary is

1. yes, cool. thanks
2. status will be "Provisional", because *we think this idea is worth implementing (We have a firm idea of what we want to do)*

I hope that I have time tomorrow morning, but Sundays always turn strange. Otherwise Monday.

nwc10:
Rebased and merged as a fast-forward.

PR #3, opened 2021-08-06
Add an RFC describing SV boolean semantics

leonerd:
While this is still in progress its ID number remains "TODO"; I will set it to the next free number just before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ These files describe the process we are trialling
 |----|-------|
 |0002|Re-implement number to string conversion|
 |0005|Everything slices|
+|0008|Stable SV Boolean Type|
 
 ### Rejected
 
--- a/rfcs/rfc0008.md
+++ b/rfcs/rfc0008.md
@@ -1,0 +1,155 @@
+# Stable SV Boolean Type
+
+## Preamble
+
+    Author:  Paul Evans <PEVANS>
+    Sponsor:
+    ID:      0008
+    Status:  Exploratory
+
+## Abstract
+
+Add to regular (defined and non-referential) scalars the ability to track whether they represent a boolean value. Values created from boolean expressions (such as `$x == $y`) will reliably remember this fact, in a way that can be introspected even if the value is stored into a variable and retrieved later.
+
+## Motivation
+
+Language interoperability concerns sometimes lead to situations where it is necessary to represent typed data - such as strings or numbers - in a way that can be reliably distinguished. Elsewhere in Perl there are attempts to add this distinction. This RFC attempts to address how to handle values of a boolean type; values that represent a simple true-or-false nature.
+
+For example, JSON and MsgPack are commonly-encountered serialisation formats that Perl can generate and parse, though in both cases while the formats themselves can represent boolean truth values distinctly from numbers or strings, perl cannot preserve that distinction. Compare this to some other languages popular at the moment:
+
+```
+$ python3 -c 'import json; print(json.dumps([1, "1", 1 == 1]))'
+[1, "1", true]
+
+$ nodejs -e 'console.log(JSON.stringify([1, "1", 1 == 1]))'
+[1,"1",true]
+
+$ perl -MJSON::MaybeUTF8=encode_json_utf8 -E 'say encode_json_utf8([1, "1", 1 == 1])'
+[1,"1",1]
+```
+
+There is a tendancy for modules to make up this shortfall with workarounds like the `JSON::PP::Boolean` type:
+
+```
+$ perl -MJSON::MaybeUTF8=decode_json_utf8 -MData::Dump=pp 
+   -E 'say pp(decode_json_utf8($ARGV[0]))' '[1,true]'
+
+[1, bless(do{\(my $o = 1)}, "JSON::PP::Boolean")]
+```
+
+Obviously such a solution is specific to JSON encoding and does not apply to, for example, message gateway between JSON and MsgPack, which would require some translation inbetween. A true in-core solution to this problem would have many benefits to interoperability of data handling between these various modules.
+
+((TODO: Add some comments about purely in-core handling as well that don't rely on serialisation))
+
+## Rationale
+
+## Specification
+
+There are two parts to this specification; the lower-level in-core details of the C code implementation that are visible to XS code; and the higher-level Perl-visible details that are made visible to Perl programs.
+
+### C-level Internals
+
+At the C level, SVs will require a macro to test whether they contain a boolean. 
+
+```
+SvIsBOOL(sv)
+```
+
+The core immortals `PL_sv_yes` and `PL_sv_no` will always respond true to `SvIsBOOL()`, as will any SV initialised by copying from them by using `sv_setsv()` and friends. The upshot here is that the result of boolean-returning ops will be `SvIsBOOL()` and this flag will remain with any copies of that value that get made - either over the arguments stack or stored in lexicals or elements of aggregate structures.
+
+These `SvIsBOOL()` values will still be subject to the usual semantics regarding macros like `SvPV` or `SvIV` - so numerically they will still be 1 or 0, and stringily they will still be "1" and "" (though see the Future Scope section below).
+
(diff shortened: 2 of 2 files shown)


iabyn:
I think this is terrible idea! We seem to be suddenly obsessed with ignoring perl's basic operating principle (that data is fluid and typeless - type interpretation being imposed by ops operating on the data) for the sake of the one use case of round trips for JSON and similar serialisers .

Such a boolean flag adds extra overhead to lots of hot places in the perl core - for example pp_add(), which currently has to handle all permutations of two args being (int/num/string/other), would have to cope with (bool/int/num/string/other) x 2

leonerd:
> I think this is terrible idea! We seem to be suddenly obsessed with ignoring perl's basic operating principle (that data is fluid and typeless - type interpretation being imposed by ops operating on the data) for the sake of the one use case of round trips for JSON and similar serialisers .

This doesn't change the semantics of any existing operations: all values ("boolean-intent" or not) can still be tested for truth, can still be numified or stringified. Nothing at all changes there.

The only thing that is new is that now there is an additional question that can be asked ("is this value boolean-intent?"), a question which answers "yes" to the result of any boolean predicate test operator, or any SV transitively initialised from such.

> Such a boolean flag adds extra overhead to lots of hot places in the perl core - for example pp_add(), which currently has to handle all permutations of two args being (int/num/string/other), would have to cope with (bool/int/num/string/other) x 2

Indeed - it turned out for that and other reasons I didn't implement it in a flag. Instead, in my branch I have a small optimisation in `sv_setsv_flags` which notices the special PV values stored by PL_sv_yes (the char array PL_Yes) and PL_sv_no (the char array PL_No) and copies those pointers, with SvLEN set to zero, into the destination. This has two upshots:

  1) We save memory and CPU time by not needing to make string copies of boolean values: e.g.

```perl
my @arr;
push @arr, 1 == 1 for 1 .. 100000;
```

will now consume less memory and take less CPU time to run.

  2) A pointer comparison on `SvPVX(sv)` is now sufficient to implement the `SvISBOOL()` test, which I have added as a function in Scalar::Util.

As far as I can tell, this optimisation is already beneficial in terms of CPU and memory savings, even if we ignore the new ability we gain by having `SvISBOOL()`

You can see my current progress at

  https://github.com/leonerd/perl5/tree/stable-bool

kraih:
> On a broader note:
> 
> > I think this is terrible idea! We seem to be suddenly obsessed
> 
> I don't know about "suddenly" there. It's been the case a long time that Perl users have cared about things like boolean roundtrips in serialisers. 

There's really nothing sudden about it. We have more cases where the community has been begging for a little "stricter" types for many many years. The old scalar value problem of characters vs bytes comes to mind, when dealing with I/O. There's a reason `utf8::is_utf8` is misused so much in CPAN modules.

marcusramberg:
Given this doesn't have a negative performance impact, it seems like a great improvement for allowing Perl to inter-op more easily with modern devops stuff, which almost always involves json or yaml wrangling.

iabyn:
On Sat, Aug 07, 2021 at 07:41:16AM -0700, Paul Evans wrote:
> As far as I can tell, this optimisation is already beneficial in terms of CPU and memory savings, even if we ignore the new ability we gain by having `SvISBOOL()`
> 
> You can see my current progress at
> 
>   https://github.com/leonerd/perl5/tree/stable-bool

The trouble with this is that it is making a shared read-only string
buffer effectively writeable. This SEGVs on your branch:

    my $x = 1;
    my $y = ($x == 1);
    $y =~ s/1/0/;

-- 
In England there is a special word which means the last sunshine
of the summer. That word is "spring".

leonerd:
> The trouble with this is that it is making a shared read-only string buffer effectively writeable. This SEGVs on your branch

Ahyes, indeed. :/ I wonder if setting the COW flag will solve this. Possibly, though it suggests there might be a possibility for other similar troubles.

In any case while I think it over I'll add some tests and a fix for this one at least.

leonerd:
OK, I think have made some progress there. I've also made a (draft) PR for the branch, so discussions on the implementation can be had over there: https://github.com/Perl/perl5/pull/19040

That leaves this thread free for the abstract intent of the idea, aside from the impl.

leonerd:
Implementation-wise, the code looks happy and solid, and so far seems to be waiting on this RFC process to continue before it gets merged.

How do we resolve this stalemate? Can we "accept" the RFC? Or failing that, do we just merge the PR that implements it into perl core and call it done?

rfcs/rfcTODO.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+SvBOOLOK(sv)
+SvBOOLOK_on(sv)
+SvBOOLOK_off(sv)
+```
+
+The core immortals `PL_sv_yes` and `PL_sv_no` will always respond true to `SvBOOLOK()`, and such a flag will be reliably copied by `sv_setsv()` and friends. The result here is that the result of boolean-returning ops will be `SvBOOLOK()` and this flag will remain with any copies of that value that get made - either over the arguments stack or stored in lexicals or elements of aggregate structures.
+
+These `SvBOOLOK()` values will still be subject to the usual semantics regarding macros like `SvPV` or `SvIV` - so numerically they will still be 1 or 0, and stringily they will still be "1" and "" (though see the Future Scope section below).
```
tonycoz: You might want to clarify here to cover @iabyn's comment, that the POK, IOK etc flags on the SV won't be changing, so code like pp_add doesn't need any changes.

rfcs/rfcTODO.md
```diff
@@ -0,0 +1,142 @@
...
+```
+$ perl -MJSON::MaybeUTF8=decode_json_utf8 -MData::Dump=pp 
+   -E 'say pp(decode_json_utf8($ARGV[0]))' '[1,true]'
+
+[1, bless(do{\(my $o = 1)}, "JSON::PP::Boolean")]
+```
+
+Obviously such a solution is specific to JSON encoding and does not apply to, for example, message gateway between JSON and MsgPack, which would require some translation inbetween. A true in-core solution to this problem would have many benefits to interoperability of data handling between these various modules.
+
+((TODO: Add some comments about purely in-core handling as well that don't rely on serialisation))
```
tonycoz, line 42: Should this TODO be updated? (or removed)

duncand:
With regard to the detail you are less sure about, the `true`/`false` keywords or `is_boolean`, I believe it would be best if they were all provided in the same place, considering that logically `true` and `false` are just nullary functions and `is_boolean` is a unary function.

If they will live with Scalar::Util then they should all be there, and one can `use Scalar::Util 'false';` just as much as they can `use Scalar::Util 'is_boolean';`.

I believe having a `use feature boolean` would be an unnecessary dedicated feature, but if you did have it then it should provide the `is_boolean` along with the `false` and `true`.

But I believe the best solution to this will also be informed by the separate discussion going on in p5p/etc about having the new generic `std`/`builtin`/etc namespace, and if that is implemented then it is where all 3/etc of these should go, and otherwise if it isn't I would say put them all in Scalar::Util.

haarg:
It seems reasonable to put true and false constants in the `std::` namespace, and leave those out until that namespace is established. Once it exists, it's probably reasonable to bring in a few different functions from Scalar::Util.

Until then, I think an `isbool` function on its own in Scalar::Util is fine.

leonerd:
@duncand Yes you're right about the `std` namespacing points. Much of the original design on this RFC predates the `std` namespace discussion, it might need adjusting in light of the outcomes from that.

kraih:
If using `true` and `false` required an import, i would always use `!!1` and `!!0` instead.

duncand:
@kraih That's fair.  What I was specifically advocating was that we do NOT have `use feature boolean;` just to get true/false.

duncand:
Actually I would argue that `false`/`true` specifically would be strong candidates for appearing automatically when one says `use 5.36;` etc.  They have a much stronger case to being available with the minimum possible boilerplate of any kind.  So better than `use feature boolean` or `use Scalar::Util` or `std::false`/`std::true`, although come to think of it `std::false` etc probably actually isn't that bad for being unambiguous.

Grinnz:
Adding it as a feature is a prerequisite for "appearing automatically when one says `use v5.36;`". Unless you want to propose an additional mechanism by which that would enable keywords, which I'm sure will be part of the function namespace discussion.

PR #5, opened 2021-10-09
source::encoding, to declare the encoding of source

rjbs:


FGasper:
I suggest `utf-8` instead of the un-hyphenated form.

Or is the idea to allow, e.g., unpaired surrogates?

rjbs:
I meant to include a note that I want to discuss, separately, strictifying the behavior of use utf8, and matching that.

johannessen:
`use source::encoding "UTF-8"` should imply that “[Perl extended UTF-8](https://metacpan.org/pod/Encode#UTF-8-vs.-utf8-vs.-UTF8)” byte sequences such as `F4908080` (U+110000) are forbidden, but those are currently allowed under `use utf8`.

However, the [documentation](https://metacpan.org/pod/utf8) mostly refers to UTF-8, so I suppose this might simply be a bug in `use utf8`.

khwilliamson:
I'm wondering what to do about non-printables.  Maybe these should be forbidden as well?  Do we want to allow an FS character, for example, literally in code, or should the program express it as an octal or hex constant?  There are various types of non-printables.; there are formats, surrogates, private use, non-characters, and above-Unicode.  What about these?

rjbs:
These are good questions.  I think the best thing we can do is probably not overcorrect from current behavior, and I'd suggest a course of action like follows -- and Karl, I'm interested in your thoughts on it:
* `use source::encoding 'utf8';` is defined as having the same effect on source decoding as `use utf8;`
* I believe we want to tighten up the meaning of `use utf8` over time; we punt this to the future.  We eliminate surrogates and private characters as well as supraästral characters.
* Further tightening the codepoint sequences allowed in a *Perl source document* as opposed to just an encoded string is another thing we can ratchet up over time.  Some may be "surely easy" to fix.  Should `CANCEL CHARACTER` be permitted in source?  If not, probably easy to remove… but I am really unsure about \Cf characters.  Eliminating bidi controls seems sensible.  Eliminating ZWJ is effectively a nonstarter.  I need to be able to put 👷🏽‍♂️ in my source code!

But all this is to say:  I believe that if we begin by making source::encoding for UTF-8 documents act like utf8.pm, we can then improve both with a clean conscience over time.

demerphq:
Fwiw I think this is wonderful improvement. In my experience dealing with a
very large number of developers new to perl "use utf8" was one of the most
common causes of deep confusion about perl strings, perl code, and their
content.

Thanks a lot!
Yves

rfcs/rfc0007.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Abstract
+
+This RFC proposes a new pragma, `source::encoding`, to indicate the encoding of
+the source document.
+
+## Motivation
+
+At present, unless in a scope in which `use utf8` has been enabled, bytes read
+from source correspond directly to the codepoints as which they are
```
khwilliamson, line 18: It doesn't make sense to me 

> as which they

PR #6, opened 2021-10-16
Namespace for Builtin Functions

leonerd:


rfcs/rfc0009.md
```diff
@@ -0,0 +1,213 @@
...
+
+How to version number the `builtin` pragma module?
+
+This becomes an especially interesting when considering the dual-life module distribution provided as a polyfill for older perls.
+
+A case can be made that its version number should match the version of perl itself for which it provides polyfill functions. Thus, code could write:
+
+```perl
+use builtin v5.40;
+# and now all the builtin:: functions from perl v5.40 are available
```
Grinnz, line 190: In addition to the bugfix concerns of tying the module version to the Perl version, this shouldn't import any functions without either a list of functions or the aforementioned bundle tags. The version would instead be useful to create a nicer error message if your version of builtin.pm is not new enough:

```perl
use builtin v1.2.3 'fribble';
# after ascertaining that this is the version that first provides the fribble function
```

So the module could have its own version scheme like any other module, and additionally document both the module version and Perl version wherein each function becomes available. I don't think this will be too confusing as long as the versions are not similar to each other, as the module version would be singularly useful for the specific mechanism above.

Grinnz:
👍 to everything here!

The namespace `builtin` is owned by GBARR on CPAN (but not in use), so this will need to be transferred to P5P for the purpose of providing the dual-life module. It last appears to have been used for https://metacpan.org/release/GBARR/builtin-0.04 in 1998 (now deleted). The easiest way to get it transferred would be to contact GBARR, and failing that contact the PAUSE admins.

EvanCarroll:
@leonerd on the mention of POSIX, how will that work without redesign? Currently when you import POSIX.pm you grow the resident size of perl substantially because [`use POSIX ()` results in massive opcode bloat.](https://stackoverflow.com/q/46738256/124486) Will the functionality required here by POSIX.pm result in a reimplementation of POSIX.xs? If so, I would love to see this kind of functionality expanded such that the rest of POSIX could be à la carte? That is to say, that you could pick any functional subset of POSIX.pm and not pay the 10x penalty in RSS.

PR #7, opened 2021-10-31
RFC 0010: kv

rjbs:


rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+
+    Author:  Ricardo Signes <rjbs@semiotic.systems>
+    Sponsor:
+    ID:      0010
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes `kv`, a new function for returning the keys and values of a
+contianer.  Among other things, this allows key/value iteration on arrays as
```
Grinnz, line 13:
```suggestion
container.  Among other things, this allows key/value iteration on arrays as
```

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+This is probably about as simple as this can get without some significant new
+addition to the language.
+
+## Specification
+
+Like `keys` or `values`, `kv` can be called with a hash expression or array
+expression.  Other types of arguments are compile time errors.
+
+In scalar context, `kv` returns the number of entries in its argument, just
+like `keys` or `values`.
```
haarg, line 68: I might have expected that it would return the size of the list it would return in list context, which would be twice the number of entries in its argument.

Also worth comparing to `->%[]` slices, which return the value of the last key/index requested.

haarg (commented):
While the name of this makes sense as an extension of keys and values, its only real purpose is when applies to arrays, where it is not returning keys but indexes. Feels a bit wrong.

A more natural idea, to me, would be to make this part of the slicing syntax. If, for example, the currently invalid syntax `->%[]` meant to return all indexes, it would accomplish the same thing this RFC is meant to cover.

rfcs/rfc0010.md
rjbs: On the other hand list `kv %hash` is the same as `%hash` so making scalar kv match that seems simpler and easy to explain.  I think this is going to be easy to explain as:

* keys does X
* values does Y
* kv does Z

All three can be applied to hashes and arrays with clear and translatable meaning.  In list context, hashes act like kv was used and arrays like values.

rjbs: Slicing for "The whole thing" was proposed in another RFC, response was neutral-to-negative, and this proposal was meant to have far fewer ramifications, capable of getting into the language to make array iteration all better at once.

leonerd: I'm with @haarg here. I really would expect the invariant that:

```perl
(scalar kv %hash) == do { my @tmp = kv %hash; scalar @tmp }
```

I.e. that `scalar kv` returns twice the number of keys or values; the total number of keys and values combined.

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+   9  11
+```
+
+## Backwards Compatibility
+
+There should be no significant backwards compatibility concerns.  `kv` will be
+imported only when requested.  Static analysis tools may need to be updated.
+
+For polyfill in older perls, I don't believe a pure-perl implementation of `kv`
+can be provided while maintaining its aliasing behavior.  This may not be true.
```
leonerd, line 114: Not a pureperl one, but an XS one can do so easily.

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+## Future Scope
+
+I believe this will be complete as is.
+
+## Rejected Ideas
+
+One suggestion was that `kv LIST` should zip together the list with would-be
+indexes.  I believe this is a complication, as it multiplies the *kind* of
+builtins surrounding container inspection.  My take is that `keys` and `values`
+don't work on lists, so this also should not.
```
leonerd, line 144: I originally made that suggestion, but now it seems neatest for `kv` to just act like a combination of `keys+values`, meshing the results of the two together and only accept the same things those components do. So I agree with you here.

A more generic "enumerate" function could always be added to `List::Util` or `builtins` or wherever, which returns `(0, LIST[0], 1, LIST[1], ...)` if required.

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Open Issues
+
+* How will this be requested?  The answer is likely: If the "builtin" namespace
+  exists when this is ready, we use that.  If not, it goes behind a `feature`
```
leonerd, line 149: Actually I disagree. `builtin` is for things that are purely functions, but as `kv` is like `keys` and `values` in that they are really basic syntax operators that operate directly on an array or hash term and not just "and ol' list", they should not be in that space.

My suggested "enumerate" or whatever it might be called, could be a builtin function. But this `kv` really is a real operator, not just a function.

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+expression.  Other types of arguments are compile time errors.
+
+In scalar context, `kv` returns the number of entries in its argument, just
+like `keys` or `values`.
+
+In list context, `kv %h` is identical in behavior to `%h` in list context.
+
+In list context, `kv @arr` returns a list twice the size of the array, zipping
+together the keys and values.  To match existing behaviors, the keys are
+mutable copies and the values are aliases.  This requires the following work:
```
leonerd, line 74: Perhaps really spell it out, that `kv @arr` will return the list

```perl
(0, $arr[0], 1, $arr[1], 2, $arr[2], ...)
```

rfcs/rfc0010.md
```diff
@@ -0,0 +1,157 @@
...
+
+Like `keys` or `values`, `kv` can be called with a hash expression or array
+expression.  Other types of arguments are compile time errors.
+
+In scalar context, `kv` returns the number of entries in its argument, just
+like `keys` or `values`.
+
+In list context, `kv %h` is identical in behavior to `%h` in list context.
+
+In list context, `kv @arr` returns a list twice the size of the array, zipping
```
oodler577, line 72: _zipping_ is a wholly different operation. What I recommend here, is to speak of `kv` as return a _tuple_ of size 2. I.e., give a `%h` or `@a`, one should expect to get each _pair_ :

* (0,1)
* (2,3)
* ...
* (`$#a-1`, `$#a`)

oodler577:
What would happen if `kv` was applied to an array with an _odd_ number of elements?

Grinnz:
I think you misunderstand the operation here. kv returns an even sized list and only half of that list are the elements of the array, the other half are the indexes.

oodler577:
> I think you misunderstand the operation here. kv returns an even sized list and only half of that list are the elements of the array, the other half are the indexes.

Can you show me an example? That doesn't sound like what `kv` _should_ do. I understood this to be a _sane_ implementation of `each`. If it does anything other than provide key/value pairs for iteration, that it shouldn't be called `kv`.

Grinnz:
Yes, it is to provide key/value pairs for iteration, where keys (as in the keys function) are the array indices and values (as in the values function) are the array elements.

leonerd:
> Can you show me an example?

```perl
my @arr = qw( zero one two );
kv @arr; # returns the list (0, "zero", 1, "one", 2, "two")
```

Grinnz:
Also, this is not "a sane implementation of each" - that is the multi-foreach syntax proposed in https://github.com/Perl/RFCs/blob/master/rfcs/rfc0001.md and implemented in https://github.com/Perl/perl5/pull/18925

This is just something that can be used alongside that to complete an "each" replacement for arrays that iterates over indices and elements.

Grinnz:
The assignment of a list of pairs between arrays and hashes is already normal behavior, as you note, and so needs no additional operator. The key/value abstraction we are discussing here is that already performed by the `keys` and `values` builtins, as well as `each`.

oodler577:
Thank. Just want to concisely say that it feels weird that `kv` is making the implicit (array indexes) explicit, but it's honoring the explicit inherent in the hash (hash keys). The only way I can make this sensible in my mind is to couch `kv` as an alternative to _hashify_ an array, taking its ordered index to be the _key_ and the element in the index slot to be the _value_. I suppose it could work if sold that way; FWIW, this is not unhelpful behavior - just a little...well I already said it :)

rfcs/rfc0010.md
oodler577: I wanted to mention after some thought, `kv` _can_ be thought of as an _unzip_ operation. That said, a true _zip_ of 2 lists would be a useful thing. See also, https://www.geeksforgeeks.org/zip-in-python/

rfcs/rfc0010.md
haarg: List lvalue functions are pretty questionable, but they do make this possible:

```
sub kv (\[%@]) :lvalue {
  Scalar::Util::reftype($_[0]) eq 'ARRAY'
    ? (wantarray ? $_[0]->%[keys $_[0]->@*] : $_[0]->@*)
    : $_[0]->%*;
}
```

rfcs/rfc0010.md
haarg: It is a good point that since `kv %hash` returns the same as `%hash` in list context, it would be reasonable for it to return the same in scalar context as well. But that also seems like optimizing for something that should never be used. `kv %hash` exists in this proposal for consistency, but is never meant to be used by anyone.

Grinnz: Perhaps it would be better if it rejected hashes outright, so we can focus on its behavior for arrays. It needs a real array or hash, so there is no benefit to accepting both.

demerphq:
On Tue, 16 Nov 2021, 01:51 Dan Book, ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In rfcs/rfc0010.md
> <https://github.com/Perl/RFCs/pull/7#discussion_r749815574>:
>
> > +```perl
> +for my ($i, $value) (kv $alpha->{beta}->[0]->@*) {
> +  say "$i == $value";
> +}
> +```
> +
> +This is probably about as simple as this can get without some significant new
> +addition to the language.
> +
> +## Specification
> +
> +Like `keys` or `values`, `kv` can be called with a hash expression or array
> +expression.  Other types of arguments are compile time errors.
> +
> +In scalar context, `kv` returns the number of entries in its argument, just
> +like `keys` or `values`.
>
> Perhaps it would be better if it rejected hashes outright, so we can focus
> on its behavior for arrays. It needs a real array or hash, so there is no
> benefit to accepting both.
>

Personally I think having a function called kv that doesn't operate on
hashes would be surprising. Why not call it iv or 'pairs' in that case?

It would however resolve an open question I would have about kv and hashes,
in that I wonder how it would interoperate with the hash iterators. Would
it reset the iterator? Just ignore it? Eg, is there a reason necessarily
that it needs to reset the iterator if it returns an even sized list of key
values? Tied hashes maybe?

Yves

>

rjbs:
Discussion from PSC today:  instead, maybe a built-in that is: "zip together a list of values with an incrementor" so you can call `xyzzy @arr` or `xyzzy list_func` and get `(0, v0, 1, v1, ...)`.

By doing it as a builtin rather than a List::Util function, we can optimize it for literal arrays.

leonerd:
I think we could easily add that as a `builtin` function. Whether we do that before or after 5.36 is a question.

rjbs:
I will write a new RFC.

PR #8, opened 2021-11-15
New RFC: command-line flag for slurping

xenu:
It's not clear to me what's the process to allocate an RFC number, so I set it to XX for now.

Pre-RFC discussion: https://www.nntp.perl.org/group/perl.perl5.porters/2021/09/msg261616.html

### DIFF
--- a/rfcs/rfc0011.md
+++ b/rfcs/rfc0011.md
@@ -1,0 +1,76 @@
+# Command-line flag for slurping
+
+## Preamble
+
+    Author: Tomasz Konojacki <me@xenu.pl>
+    Sponsor:
+    ID: 0011
+    Status: Draft
+
+## Abstract
+
+Introduce a new command-line flag for `perl`, `-g`, which sets `$/` to `undef`,
+and thus enables slurp mode. It is a simpler alias for `-0777`.
+
+## Motivation
+
+Slurping (i.e. reading a whole file at once, instead of line by line) is a very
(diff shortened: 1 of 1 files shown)


neilb:
You get **0011**.

xenu:
Updated.

rfcs/rfc0011.md
```diff
@@ -0,0 +1,76 @@
...
+
+    Author: Tomasz Konojacki <me@xenu.pl>
+    Sponsor:
+    ID: 0011
+    Status: Draft
+
+## Abstract
+
+Introduce a new command-line flag for `perl`, `-g`, which sets `$/` to `undef`,
+and thus enables slurp mode.
```
oodler577, line 13: It would be helpful to note this is to create a flag that is equivalent to `-0777` (only mented below).

oodler577:
Any support to changing this from `-g` (common for debugging via `gcc`) with `-G`? The capitalization would indicate something special about it (it is a _special case_ of `-0`) and as far as I know, it's not used as a _defacto_ flag for anything. Thank you for writing up the RFC and for the consideration of this suggestion.

xenu:
Collisions with other programs' single-letter flags are unavoidable. I don't think it's something worth worrying about.

rfcs/rfc0011.md
xenu: Done.

leonerd (approved):
LGTM

PR #9, opened 2021-11-23
RFCxxxx: Support overloaded objects in join(), substr() builtins

book:
Pre-RFC discussion: https://www.nntp.perl.org/group/perl.perl5.porters/2021/11/msg261831.html


rfcs/rfcXXXX.md
```diff
@@ -0,0 +1,154 @@
...
+
+The current behaviour is inconsistent and confusing.
+
+There is at least one CPAN module which works around the `join`
+deficiency, (the [`join` function of
+`String::Tagged`](https://metacpan.org/pod/String::Tagged#join), but not
+universally.
+
+The CPAN module
+`[overload::substr](https://metacpan.org/pod/overload::substr)`) is a
```
ilmari, line 36:
```suggestion
[`overload::substr`](https://metacpan.org/pod/overload::substr) is a
```

rfcs/rfcXXXX.md
```diff
@@ -0,0 +1,154 @@
...
+
+If code relies upon the `join` function to convert objects to plain
+strings, this would break that code, thus a feature flag seems
+necessary.
+
+There is no need autogenerate an implementation for `substr` if it is
+missing, as the stringification will happen as it does now.
+
+Modules which implement their own `join` would need to add
+version-conditional logic to work with different versions of Perl.
```
ilmari, line 79: Surely they don't _need_ to, unless they enable the feature flag (which would presumably be lexical)?

However, they can (conditionally) take advantage of the built-in `join` respecting overloading, and delegate to that when available.

rfcs/rfcXXXX.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Security Implications
+
+We do not yet foresee security issues. Guidance is welcome.
+
+## Examples
+
+For `join`:
+
+    ```perl
```
ilmari, line 92: Use either indentation or code fences for verbatim blocks, not both. Code fences are better, since they let you specify the language.

ilmari (changes requested):
:+1: for the concept!

leonerd:
LGTM

By my count the next RFC number would be 0010.

neilb:
Can you set the number to 0013 please @book ?

rfcs/rfc0013.md
```diff
@@ -0,0 +1,164 @@
+# Support overloaded objects in join(), substr() builtins
+
+## Preamble
+
+    Authors: Eric Herman <eric@freesa.org>, Philippe Bruhat <book@cpan.org>
+    Sponsor: Paul Evans <leonerd@leonerd.org.uk>
+    ID:
```
leonerd, line 7: This should be 0013

PR #10, opened 2022-01-25
0012 configure taint

neilb:
This is an RFC to add a Configure question on whether you want to build Perl with taint support, or without.

See [Pre-RFC](https://www.nntp.perl.org/group/perl.perl5.porters/2021/08/msg261250.html) discussion on p5p.

### DIFF
--- a/rfcs/rfc0012.md
+++ b/rfcs/rfc0012.md
@@ -1,0 +1,160 @@
+# Configure option for not including taint support
+
+## Preamble
+
+    Author:  Neil Bowers <neilb@cpan.org>
+    Sponsor:
+    ID:      0012
+    Status:  Draft
+
+
+## Abstract
+
+Add a Configure option for deciding whether to include taint support
+in this build of Perl.
+This is essentially a Configure wrapper
+around the existing `-DNO_TAINT_SUPPORT` compile flag.
+The question would be worded as
+
+> "Do you want to build Perl without taint support? [n]",
+
+so the default of "No" means that Perl will build as previously.
+
+## Motivation
+
+* Support for taint adds a runtime overhead of roughly 10%.
+* Adding a Configure option makes it easier for people to build a
+  Perl without taint support.
+* You will then be able to query %Config to determine whether your
+  perl was built with taint support.
+
+## Rationale
+
+* Anecdotally, a very high percentage of Perl developers never use
+  the taint features, yet we're all taking the performance hit.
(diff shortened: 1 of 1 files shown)


rfcs/rfc0012.md
```diff
@@ -0,0 +1,176 @@
...
+
+### What do we mean by "no Taint support"?
+
+There are currently two compilation options:
+
+1. `SILENT_NO_TAINT_SUPPORT`: no taint support, but you can still call `-T`,
+   it just won't do anything.
+2. `NO_TAINT_SUPPORT`: no taint support, trying to call `-T` is fatal.
+
+I propose that the Configure option sets `NO_TAINT_SUPPORT`.
```
Leont, line 58: I vehemently disagree with this. A [quick CPAN grep](https://grep.metacpan.org/search?size=20&q=%5E%23%5B%5E%5Cn%5D*perl%5B%5E%5Cn%5D-T&qd=&qft=.t) stopped looking for more test files that would be broken after the first 2000.

Taint is a stricture, much like `use strict`. Just like programs shouldn't fail when you remove `use strict`, they shouldn't either when you remove taintedness. Making it fail loudly is a fine option for research purposes, but not something that will result in a usable CPAN in any near future.

Tux: I don't see this as a show-stopper: just ignore the `-T` from those lines. My bet is that lots of scripts added it from a template or just because "they were used to do so".
Is there any other reason to oppose?

Leont: > I don't see this as a show-stopper: just ignore the -T from those lines.

That's exactly what `SILENT_NO_TAINT_SUPPORT` does, that was my point.

> My bet is that lots of scripts added it from a template or just because "they were used to do so".

Probably. I have only once used it on purpose in a test (and it's not coincidence it was named `taint.t`).

neilb:
So we have two options:

1. Make it fatal, in which case lots of dists would fail to install, because they needlessly have -T on tests
2. Make it silently non-fatal, in which case all the needless -T tests would probably pass, and some tests might fail, and it depends on how the tests are written on whether it will be obvious to the person installing why they failed.

Not really much of a choice – pragmatically we have to go with #2.

neilb:
I've just pushed an updated version, which proposes `-DSILENT_NO_TAINT_SUPPORT` rather than `-DNO_TAINT_SUPPORT`, and adds a couple of points to the Rejected Ideas section.

PR #12, opened 2022-02-19
rfc0016: indexed

rjbs:


haarg:
There's a number of places in this that still say `kv` rather than `indexed`.

rfc0016.md
```diff
@@ -0,0 +1,124 @@
...
+
+```perl
+for my ($i, $value) ($alpha->{beta}->[0]->%[ keys $alpha->{beta}->[0]->@* ]) {
+  say "$i == $value";
+}
+```
+
+This is pretty bad.
+
+With `kv`, we write this:
```
Leont, line 44: I presume this `kv` should also be `indexed`

oodler577 (approved):
This is well done. I hope this (and n-at-a-time foreach) opens the door for a complementary "interleaved"/"zip"/"meshed" keyword in core rather than in List::Utils, etc.

neilb:
> In list context, `indexed LIST` evalutes to a list twice the size of the array

I think that should be "twice the size of LIST".

PR #13, opened 2022-03-08
draft rfc for scalar type functions

haarg:


rfcs/rfc0017.md
```diff
@@ -0,0 +1,149 @@
...
+
+    Author:  Graham Knop <haarg@haarg.org>
+    Sponsor:
+    ID:      0017
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes adding new builtin functions created_as_string and
+created_as_number for checking the original form of non-scalar references.
```
demerphq, line 13: Shouldn't  it be created_as_integer and also created_as_float? Or are we not going to differentiate between NV and IV?

haarg: It is not yet possible to make that distinction. It is included in the "Future Scope" section though. I think a check for "number" is still valuable.

rfcs/rfc0017.md
```diff
@@ -0,0 +1,149 @@
...
+  my $number1 = created_as_number $value1; # false
+
+  my $value2 = 1;
+  my $string2 = created_as_string $value2; # false
+  my $number2 = created_as_number $value2; # true
+
+  my $value3 = "1";
+  my $used_as_number3 = 0+$value3;
+  my $string3 = created_as_string $value2; # true
+  my $number3 = created_as_number $value2; # false
```
johannessen, line 105: `$value2` → `$value3`

rfcs/rfc0017.md
```diff
@@ -0,0 +1,149 @@
...
+
+  my $value4 = 1;
+  my $used_as_string4 = "$value4";
+  my $string4 = created_as_string $value2; # false
+  my $number4 = created_as_number $value2; # true
```
johannessen, line 110: `$value2` → `$value4`

rfcs/rfc0017.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Prototype Implementation
+
+None.
+
+## Future Scope
+
+It is possible that in the future, the NOK or IOK flags will only be set for
+values created as numbers. That would mean the implementation of these
+functions may change, but should not impact the results returned.
```
leonerd, line 120: One slightly awkward upshot is that in the current implementation, values that start off as strings can still gain the NOK or IOK flags. Because of that, it means we can't distinguish created-as-strings from genuine dualvars (either core variables like `$!` or the result of `Scalar::Util::dualvar`).

There's not much we can do about that at present, but it may be something we want to take note of when documenting how `created_as_string` works at the moment. A future version may not be confused by dualvars, but currently both of these will be true:

```perl
created_as_string $!;
created_as_string dualvar(123, "four-five-six");
```

rfcs/rfc0017.md
exodist: Would it maybe make more sense to implement something like `%flags = scalar_flags($foo)` or `scalar_props($foo)` that returns all the interesting bits you do not normally see, including the things these functions look at, plus some summary keys for `created_as_xxx`? Then document the gotchas and what each thing may or may not mean?

Or are these specific questions really important to answer while the other things contained under the hood of a scalar are useless at the perl-code level?

haarg: That kind of information is already available via B, but it is hard to interpret correctly. And the specific behavior of the internal flags may change over time. The fact that they can change without it being a backwards compatibility break is in fact why this RFC is viable at all. Making these flags more visible would often be misleading, as they will change based on read operations. The changes in perl 5.36 make it possible to use the flags to accurately check the type, but only for the specific case of string vs number.

rfcs/rfc0017.md
```diff
@@ -0,0 +1,149 @@
...
+  my $number3 = created_as_number $value3; # false
+
+  my $value4 = 1;
+  my $used_as_string4 = "$value4";
+  my $string4 = created_as_string $value4; # false
+  my $number4 = created_as_number $value4; # true
+
+## Prototype Implementation
+
+None.
```
leonerd, line 114: I've started a (currently-draft) core PR for this at https://github.com/Perl/perl5/pull/19524

leonerd (approved):
LGTM

PR #14, opened 2022-04-03
New RFC: "yield_true" at end of compilation

Ovid:
This RFC is to eliminate the need for a magic true value (usually a hard-coded trailing `1`) being required in code we use.

I guess as to the RFC number. If there is a process for assigning them, I've missed it. My apologies.

Better names than `yield_true` are welcome. I was tempted to go with `use feature 'neo';` because "Neo is the One", but it's  terrible name, even though it amuses me :)

### DIFF
--- a/rfcs/rfc0018.md
+++ b/rfcs/rfc0018.md
@@ -1,0 +1,69 @@
+# A feature to automatically "yield true" at the end of a file
+
+## Preamble
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      0018
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes a `yield_true` feature. When used, the current Perl _file_
+containing `use feature 'yield_true'l` will automatically return a true value
+after successful compilation, eliminating the neeed for a "1" (or other true
+value) at the end of the file.
+
+## Motivation
+
+Eliminate the need for a true value at the end of a Perl file.
+
+## Rationale
+
+There's no need to have a true value be hard-coded in our files that we
+`use`. Further, newer programmers can get confused because sometimes code
+_doesn't_ end with a true value but nonetheless compiles just fine because
+_something_ in the code returned a true value and the code compiles as a
+side-effect.
+
+
+## Specification
+
+    use feature 'yield_true';
+
+Code using the above does not need to return a magic true value when compiled.
+
+## Backwards Compatibility
+
+There are no compatibility concerns I'm aware of.
+
+## Security Implications
+
+None expected.
+
+## Examples
+
+None. See above.
+
+## Prototype Implementation
+
+None.
+
+## Future Scope
+
+I believe this is complete.
+
+## Rejected Ideas
+
+None that I'm aware of.
+
+## Open Issues
+
+None that I'm aware of.
+
+## Copyright
+
+Copyright (C) 2022, Curtis "Ovid" Poe
+
+This document and code and documentation within it may be used, redistributed
+and/or modified under the same terms as Perl itself.


haarg:
It is possible for a file to explicitly return a false value. I'd expect that would just be ignored with this feature enabled, but it's probably worth specifying.

rfcs/rfc0018.md
```diff
@@ -0,0 +1,69 @@
...
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      0018
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes a `yield_true` feature. When used, the current Perl _file_
+containing `use feature 'yield_true'l` will automatically return a true value
```
leonerd, line 13: Typo, you probably didn't mean the "1" in here

rfcs/rfc0018.md
```diff
@@ -0,0 +1,69 @@
...
+
+None. See above.
+
+## Prototype Implementation
+
+None.
+
+## Future Scope
+
+I believe this is complete.
```
leonerd, line 54: The intention of having it be a named feature, is that a future version bundle will include it, so you'll get it automatically with, perhaps,

```perl
use v5.40;
```

haarg:
A prototype implementation exists at https://metacpan.org/pod/true

rfcs/rfc0018.md
```diff
@@ -0,0 +1,69 @@
...
+_doesn't_ end with a true value but nonetheless compiles just fine because
+_something_ in the code returned a true value and the code compiles as a
+side-effect.
+
+
+## Specification
+
+    use feature 'yield_true';
+
+Code using the above does not need to return a magic true value when compiled.
```
leonerd, line 34: Something to think about is that while *most* modules just put "1" or some other dummy true value, there are some modules that make use of the return value, because it becomes the value returned by a `require`. In particular, it is not uncommon to yield the package name itself, when that is an object class, so you can do things like

```perl
my $obj = (require Some::Object::Class)->new;
```

I don't think we want to complicate this feature idea by suggesting we'd allow customisation of the true value being returned, but perhaps if we're returning some value anyway, the string form of the package name is just as good as any, and makes this use-case still work. (And anyone who writes `package 0` deserves what they get :P)

Corion:
I'm mildly opposed to simply returning `1` because I still entertain the idea of actually (optionally) returning an object. Returning the package name is a good thing though. The return value of `require My::Example;` should always return the package name, even on the second invocation then:

```
for (1..2) {
    my $factory = require My::Example;
    my $handler = $factory->new($_);
}
```

haarg:
Changing the return value of require seems like a separate issue from what this RFC wants to address.

If you wanted require to always return the same value, and for that value to come from the file, you need a new location to store these values. This would probably mean a new superglobal to go along with %INC. And it would usually be useless because most modules return 1. I don't think this is a very good idea.

If you wanted require to always return the package name, it's a separate issue from this RFC because it's ignoring the return value from the file anyway. It also presents a problem because require doesn't actually take package names. It takes file path fragments. Foo::Bar is translated to "Foo/Bar.pm" at parse time. This would then need to be converted back to a package name, or do something else. I don't think this is a good idea either.

Instead, it's probably best addressed with a `builtin::load` or similar routine that accepts a package as a string. This has been discussed in the past, and solves other problems. Module::Runtime has a use_module function that behaves like this, returning the package name.

leonerd:
@haarg makes some good points about returning odd values, so yes probably best to stick to just "1" (or more likely the new "true" value)

However, in saying the word "return", does raise an interesting point. Most modules currently do not *return* a true value, they simply yield one by being the final expression in the block. It is possible to explicitly `return` a value. Presumably this is good enough to override whatever default-true value this feature would propose.

```perl
package Some::Module v1.23;
use feature 'yield_true';

return "a different value";
```

haarg:
Would returning a false value also override the true value? It feels a bit strange to me. Maybe an explicit return in the context of `yield_true` should be an error.

Ovid:
I accidentally wrote this in `master` instead of a branch. Closing this in favor of https://github.com/Perl/RFCs/pull/16

Issues raised here have been addressed. Sorry 'bout that.

demerphq:
On Sun, 3 Apr 2022, 17:44 Graham Knop, ***@***.***> wrote:

> It is possible for a file to explicitly return a false value. I'd expect
> that would just be ignored with this feature enabled, but it's probably
> worth specifying.
>

That would be bad.

There is a difference between returning false and dieing during require.

Returning false allows the code to be required again, possibly after fixing
the cause of the false return, dieing does not allow the file to be loaded
again. (Fixing part of this, where death is signified by a specific undef,
is on my todo list.)

Therefore this feature must distinguish and deal with modules which
explicitly return false.

Yves

>

haarg:
I'm not convinced that is a behavior that needs to be preserved under this feature. Allowing a module to be required multiple times is mostly a surprising behavior, and anything expecting to handle that programmatically would need to either be working with known modules or need some external way of communicating what is needed to load the module. And if the caller knows it can attempt loading the module multiple times, it can remove the `%INC` entry to allow it. Or, the module in question can disable (or leave disabled) this feature.

PR #15, opened 2022-04-03
RFC: Add :export attribute RFC

Ovid:
This RFC (no number) is to provide an `:export` attribute to Perl to make exporting both easier and lexical.


rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+        sub bar  :export            {...} # @EXPORT_OK
+        sub baz  :export(:strings)  {...} # %EXPORT_TAGS
+        sub bay  :export(:strings)  {...} # %EXPORT_TAGS
+        sub quux :export(:ALWAYS)   {...} # @EXPORT
+        sub whee                    {...} # not exported
+    }
+
+If an `import()` method is defined, it's possible that it might have special
+importing functionality provided, but we will _still_ allowing importing of
+functions with `:export` tags. We pass the import arguments unchanged to the
```
haarg, line 172: I'd disagree with this. I don't think there should be a new magic layer before `import`.

Instead, I'd say that the exporting should be implemented in UNIVERSAL::import, and if a module wants to use `:export` and have its own import, it should call UNIVERSAL::import in its import method with the appropriate parameters.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+
+        sub bar  {...}
+        sub baz  {...}
+        sub bay  {...}
+        sub quux {...}
+        sub whee {...}
+    }
+
+
+Imported functions do not remain in the namespace, removing the need for
```
haarg, line 83: Should be explicit that the imported functions are created as lexical subs.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+we do not need a feature guard. Instead, we can simply `use v5.XX` to pull in
+the new feature.
+
+    package My::Module {
+        use v5.38;
+
+        sub bar  :export            {...} # @EXPORT_OK
+        sub baz  :export(:strings)  {...} # %EXPORT_TAGS
+        sub bay  :export(:strings)  {...} # %EXPORT_TAGS
+        sub quux :export(:ALWAYS)   {...} # @EXPORT
```
haarg, line 55: `:ALWAYS` is probably not the best name, because the entries are not _always_ imported, but imported by default when nothing is specified.

Exporter allows importing the `:DEFAULT` tag, which is the entries in `@EXPORT`. Reusing that name is probably an improvement on `:ALWAYS`, and it could also be allowed as an importable tag for parity with Exporter.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+As a counter-argument, it's unlikely someone will be using `:export`
+attributes, so we could simply filter out all subnames so they don't populate
+`@args`, or filter out all subnames tagged with `:export`.
+
+For the above, that means that these are problematic:
+
+    use My::Module ':srtings';
+    use My::Module 'whee';
+
+For `:srtings`, that's almost certainly misspelled. For `whee`, that's not
```
leonerd, line 186: Typo "strings"

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+
+    use My::Module 'whee'; # fatal error
+
+## Prototype Implementation
+
+None that I'm aware of, though
+[Perl6::Export::Attrs](https://metacpan.org/pod/Perl6::Export::Attrs) seems
+the closest.
+
+## Future Scope
```
leonerd, line 229: Another possible future idea: importing as another name. Since they're lexical imports and not added to the externally-visible namespace of the module that does the importing, it should be safe to rename them. Maybe support some notation at import time, where an importer can import a function as a different name to avoid clashes or for general clarity.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+* [Exporter::Simple](https://metacpan.org/pod/Exporter::Simple)
+* [Perl6::Export](https://metacpan.org/pod/Perl6::Export) (not really attributes)
+* [Perl6::Export::Attrs](https://metacpan.org/pod/Perl6::Export::Attrs)
+
+And so on.
+
+## Specification
+
+Because lower-case attribute names are reserved for Perl's future expansion,
+we do not need a feature guard. Instead, we can simply `use v5.XX` to pull in
```
leonerd, line 46: Actually, they don't even need this. We can add a new `:export` attribute that becomes visible by default without needing to do anything to request it. No harm in the user code doing `use v5.38` but also no need to.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
+# Built-in lexical exporting
+
+## Preamble
+
+    Author:  Curtis Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      XX
+    Status:  Draft
```
johannessen, line 8: I suggest to follow [the proposed self-assigned ID scheme](https://github.com/Perl/RFCs/blob/main/docs/future.md#rfc-ids-and-how-to-self-assign) as long as this remains a draft: `OVID-XXXX`

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+
+And so on.
+
+## Specification
+
+Because lower-case attribute names are reserved for Perl's future expansion,
+we do not need a feature guard. Instead, we can simply `use v5.XX` to pull in
+the new feature.
+
+    package My::Module {
```
leonerd, line 49: Small note on markdown formatting: I usually like to use the triple-backtick guard for these, so I can add the word "perl" and get better syntax highlight.

    ```perl
    code here ...
    ```

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+        use v5.38;
+
+        sub bar  :export            {...} # @EXPORT_OK
+        sub baz  :export(:strings)  {...} # %EXPORT_TAGS
+        sub bay  :export(:strings)  {...} # %EXPORT_TAGS
+        sub quux :export(:ALWAYS)   {...} # @EXPORT
+        sub whee                    {...} # not exported
+    }
+
+The above can be written as this:
```
leonerd, line 59: I don't think I'd write "can be" here - it's not strictly equivalent. The main point of these `:export` attributes is that it's imported lexically by the caller, not symbolically. I'd perhaps word this:

"The above behaves similar to if it had been written as:"

rfcs/rfc00XX.md
leonerd, line 83: Better wording here might be:

"The requested functions are imported lexically, rather than appearing in the package symbol table. This means that code within that lexical scope can see and call them as normal, but they are not visible from outside - such as from other callers, or as object or class methods. This removes the need for `namespace::clean` and related modules."

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+    use My::Module ':srtings';
+    use My::Module 'whee';
+
+For `:srtings`, that's almost certainly misspelled. For `whee`, that's not
+supposed to be exported, but perhaps the module author wants to allow that in
+the argument list.
+
+To side-step this, we could create an alternate to the `import()` method,
+perhaps as something analogous to a phaser, but taking arguments, or a new
+`IMPORT` method. Bike-shedding may now commence!
```
leonerd, line 192: I admit I am somewhat swayed by the idea of an `IMPORT` phaser, see below.

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Future Scope
+
+We might want to allow variables and constants to be exported. We might always
+want export tags to contain other export tags to allow "grouping" of groups.
+
+The 'is not exported' error message could list what is actually exported by
+the module.
+
+An introspection API for procedural code (similar to a MOP), could be
```
leonerd, line 237: Some kind of API would be nice, yes. Combined with the `IMPORT` phaser idea, I suspect a nice way to implement this would be to just make something initially that works, and then create some API functions (perhaps in the `builtin::` space) and state that the default behaviour is equivalent to if you had written some standard example as an IMPORT phaser, using those functions. That way, if people want to understand what the behaviour is, they can read that example code. If they want to implement some slight variation on the theme, they can copy-paste it then edit as appropriate.

In effect, we'd say that the default provided implementation is *as if* the user had written that example block (though, likely internally implemented in a more efficient manner).

rfcs/rfc00XX.md
Ovid: I admit that your suggestion is much better than mine. However, I'm concerned and I have questions.

First, all procedural modules _inherit_ from `UNIVERSAL`. This is kind of a weird thing which infects _all_ procedural modules. So what happens if a module doesn't use `:export` and calls the superclass `import`? Are there bugs there if the import list mentions a function name? I don't see this as being a problem _per se_, but relying on inheritance for importing is something I would prefer to avoid.

Comments welcome. See @leonerd's comments about an `IMPORT` phaser below.

rfcs/rfc00XX.md
Ovid: I like this and will think about it. I would like to see more concrete examples and suggestions, so I won't write anything up yet.

Ovid:
Per [johannessen](https://github.com/johannessen)'s comments, this has been renamed to [rfcs/rfcOVID-0019.md](https://github.com/Perl/RFCs/pull/15/files#diff-4e7396e28058c121275c2d9a34a2508cc32dbd192855dc58153d8160aa1e7707).

rfcs/rfc00XX.md
haarg: I don't see the advantage of a phaser over just a method, as is currently done. A phaser accepting arguments is rather strange to start with. And then it needs some way to process some of them and allow others to be processed by other phasers. None of this is an issue for a method.

rfcs/rfc00XX.md
haarg: Regarding introspection, I kind of like the idea of just reusing the current variables Exporter uses.

rjbs:
I stand by [what I said](http://nntp.perl.org/group/perl.perl5.porters/262791) to a previous iteration of this on p5p:  I think you're skipping a layer.  We should be providing:

```perl
lexporter->export($name => $code);
```

…to export the coderef into currently-compiling scope as a lexical subroutine.  From this, I believe everything else above could be implemented in Perl, as could other export APIs.

demerphq:
On Sun, 3 Apr 2022 at 18:32, Graham Knop ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In rfcs/rfc00XX.md
> <https://github.com/Perl/RFCs/pull/15#discussion_r841247293>:
>
> > +
> +## Prototype Implementation
> +
> +None that I'm aware of, though
> +[Perl6::Export::Attrs](https://metacpan.org/pod/Perl6::Export::Attrs) seems
> +the closest.
> +
> +## Future Scope
> +
> +We might want to allow variables and constants to be exported. We might always
> +want export tags to contain other export tags to allow "grouping" of groups.
> +
> +The 'is not exported' error message could list what is actually exported by
> +the module.
> +
> +An introspection API for procedural code (similar to a MOP), could be
>
> Regarding introspection, I kind of like the idea of just reusing the
> current variables Exporter uses.
>
 That was in my suggestion in the pre-rfc also, and the consensus on #p5p
at the time it was discussed.

Yves

-- 
perl -Mre=debug -e "/just|another|perl|hacker/"

Ovid:
@demerphq: this was part of a larger pre-RFC that I had made. The idea was rejected, but I was told that "yield true" and the export syntax both should be considered. Later, [Paul Evans specifically said these two ideas should be written up as RFCs](https://www.nntp.perl.org/group/perl.perl5.porters/2022/04/msg263440.html), so I've done so. There's been a lot of discussion, so that was easy to miss.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,281 @@
...
+
+* [Exporter::Simple](https://metacpan.org/pod/Exporter::Simple)
+* [Perl6::Export](https://metacpan.org/pod/Perl6::Export) (not really attributes)
+* [Perl6::Export::Attrs](https://metacpan.org/pod/Perl6::Export::Attrs)
+
+And so on.
+
+## Specification
+
+Because lower-case attribute names are reserved for Perl's future expansion,
```
happy-barney, line 45: It's always good to have a human readable string to identify, why statement is there.

for example, it's better to do `use feature 'say';` then `use v5.10` - simply because reader knows what you want to do.

Therefor every new feature should have at least string representation (related module: [Syntax::Construct](https://metacpan.org/pod/Syntax::Construct))

iabyn:
From reading this I don't have the foggiest idea what the :export atrribute actually does? At compile time, does it trigger an immediate insert of an alias of a lexical sub in the caller's lexical scope? Or does it just mark some internal state on the CV for future examination? Or add to some state within the current package? Where this state is later used by a call to import() to decide what lexical aliases to create in the caller? Or what? E.g. what should be the effect of 'eval q{sub foo :export {...}}' ?

haarg:
`sub foo :export { ... }` is roughly analogous to `sub foo { ... } push @EXPORT_OK, 'foo';` for a module using Exporter. The RFC as is doesn't include any details on implementation.

demerphq:
On Wed, 6 Apr 2022 at 14:13, Graham Knop ***@***.***> wrote:

> sub foo :export { ... } is roughly analogous to sub foo { ... } push
> @EXPORT_OK, 'foo'; for a module using Exporter. The RFC as is doesn't
> include any details on implementation.
>
At least that is what I proposed in pre-rfc for making UNIVERSAL::import()
handle this.

Yves

-- 
perl -Mre=debug -e "/just|another|perl|hacker/"

Ovid:
@haarg @demerphq Both of you have mentioned the lack of implementation details here. Given that my previous RFC work focused on the user-experience and not the implementation, I thought that was OK. Is it not?

leonerd:
The entire point of the RFC process is to discuss the "user-facing experience", and design something that people want to use, without particular reference to how it's implemented internally. I think it's fine to omit such details, provided there's a general feeling that "it will be possible to implement in some manner" - which I feel this is.

demerphq:
On Thu, 7 Apr 2022 at 12:00, Paul Evans ***@***.***> wrote:

> The entire point of the RFC process is to discuss the "user-facing
> experience", and design something that people want to use, without
> particular reference to how it's implemented internally. I think it's fine
> to omit such details, provided there's a general feeling that "it will be
> possible to implement in some manner" - which I feel this is.
>
I think there is a subtle difference between what it will do and how it
will do it, and ovid has not really explained what it will do. Does it
interact with @EXPORT? Does it not? What global state will this change? How
will its effect be introspected? All of those are part of the user facing
experience.

In my pre-rfc suggesting basically exactly the same thing, which was
rejected, I explained both what it would do /and/ how.  I agree in an RFC
the how shouldnt be necessary, provided its plausible to a practitioner of
the arts, but I also dont think that Ovid has not fully explained what it
will /do/.

I think that Pre-RFC should be reconsidered and merged with this RFC.

I think for the long term good of the community in terms of developer
experience we should have a single, *well defined* and standard way to
export functions, and we should move away from the TIMTOWTDI philosophy
which just means that Perl does not play nicely with modern development
tools, which is /definitely/ one of the reasons we are losing market share.
It is absolutely one of the reasons at least one major company is moving
away from using Perl.

cheers,
Yves

-- 
perl -Mre=debug -e "/just|another|perl|hacker/"

happy-barney:
I think there is a subtle difference between what it will do and how it
> will do it, and ovid has not really explained what it will do. Does it
> interact with @EXPORT? Does it not? What global state will this change? How
> will its effect be introspected? All of those are part of the user facing
> experience.
>
>
Since this thread started I processed many modules using EXPORT* variables
and my opinion now is:

- new import/export mechanism should treat its own definition variables
(hidden)
- writing to EXPORT variables should only extend those definitions (never
overload, including tags)
- reading EXPORT variables should return (read only) content of definition
variables

imho best place for describing this is test file - and so pre-RFCs / RFCs
should not be
only text files, but should also provide tests testing proposed behaviour.

(and there lies another topic: how to write readable tests)

Best regads,
Brano

rfcs/rfc00XX.md
```diff
@@ -0,0 +1,257 @@
...
+        sub bar  {...}
+        sub baz  {...}
+        sub bay  {...}
+        sub quux {...}
+        sub whee {...}
+    }
+
+
+Imported functions do not remain in the namespace, removing the need for
+namespace::clean and relateed modules
```
tonycoz: related

rfcs/rfc00XX.md
Ovid: @haarg I've pushed an updated version. In particular, I'm concerned about this part because I think I've missed something.

---
If an `import()` method is defined in the package, it will need to call `UNIVERSAL::import()` to have the lexical importing handled correctly.

```perl
sub import ($class, @args) {
    my ( $import, $mine ) = separate_args(@args);
    # do stuff with $mine
    UNIVERSAL::import($import->@*);
}
```

For the above, how do we tell `UNIVERSAL::import` which class it's importing into? I suppose we could do `local @_ = $import->@*; goto &UNIVERSAL::import`, but that's finicky, it's going to break, and we're trying to discourage use of
`@_` in signatured subs. I expect there's some new work done in Perl which I'm missing here.

demerphq:
@leonerd wrote:
>The entire point of the RFC process is to discuss the "user-facing experience", and design something that people want to use, without particular reference to how it's implemented internally. I think it's fine to omit such details, provided there's a general feeling that "it will be possible to implement in some manner" - which I feel this is.

I rather strongly don't agree. RFC's that aren't grounded in reality of implementation miss important details and bake into the RFC things that we can't deliver or are broken or inconsistent with other code. (Eg the "where do the attributes go" fiasco that happened when function signatures were added.) They also increase the chance that the RFC never gets implemented because its choices are so difficult to implement. 

When I proposed an :export attribute I suggested that we combine it with baking in a UNIVERSAL::import which can manage the desired semantics.  That would allow us to do better with an existing set of bugs related to case insensitive file systems and provide a default bake in "exporter" mechanism at the same time. When this came up the discussion of what that default import would do. Would the exported subs be introspectable? How would it interface with existing mechanisms like Exporter. What would happen if someone puts :export and use Exporter into the same package? What about other exporters? What happens if import is overloaded and someone uses :export?

So there are a lot of questions to answer that need to be answered at an implementation level. As far as I can tell there is nothing in this RFC that cannot be done in pure perl right now. So this RFC should be implemented as a module first, then converted to live in core as XS code in unversal.c once the semantics are ironed out.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,329 @@
...
+
+```perl
+sub import ($class, @args) {
+    my ( $import, $mine ) = separate_args(@args);
+    # do stuff with $mine
+    UNIVERSAL::import($import->@*);
+}
+```
+
+For the above, how do we tell `UNIVERSAL::import` which class its importing
```
haarg, line 254: The correct way to call this would be `$class->UNIVERSAL::import(...)`. This is the same as wrapping any other method. Since this is doing lexical exports, there is no package being imported into. `caller` does not need to be maintained or communicated.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,329 @@
...
+### Interface
+
+The various incantations of `:export` should populate the respective
+`@EXPORT`, `@EXPORT_OK`, and `%EXPORT_TAGS` package variables to maintain a
+consistent API with the `Exporter` module.
+
+The grammar for the tag is:
+
+    <export>  ::= ':export' [ <options> ]
+    <options> ::=  '(' ( 'DEFAULT' | <tags> ) ')'
```
haarg, line 110: This grammar doesn't allow listing an export in a tag and in DEFAULT. That should be allowed.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,329 @@
...
+procedural code. That pre-RFC was rejected, but I was asked to put this into
+its own RFC.
+
+## Open Issues
+
+None?
+
+## Copyright
+
+Copyright (C) 20222-23, Curtis "Ovid" Poe
```
haarg, line 326: Hopefully we don't have to wait quite that long.

rfcs/rfcOVID-0019.md
Ovid: Fixed.

rfcs/rfcOVID-0019.md
Ovid: Fixed.

rfcs/rfcOVID-0019.md
Ovid: Fixed (oops).

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,329 @@
...
+
+use My::Package 'bar';     # imports bar() and quux
+
+use My::Package ':strings' # imports baz, bay, and quux
+
+use My::Package;           # imports quux
+
+use My::Package ();        # imports nothing
+
+use My::Package qw/bar :strings -quux/; # don't import quux
```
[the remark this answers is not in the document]
Ovid: Fixed.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,329 @@
...
+
+Any function that has a tag or tags (such as `strings` in the above example),
+can also be imported directly by name.
+
+### Importing
+
+* `use Mod;` Import only functions declared with `:export(DEFAULT)`
+* `use Mod qw(foo bar baz)` Import functions iff declared with `:export`
+* `use Mod ':strings';` Import only functions tagged with `strings`
+* `use Mod ':all';` All function with any form of `:export` tag are imported
```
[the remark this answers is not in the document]
Ovid: Fixed.

I think all of your current concerns are addressed. Thank you for your great feedback!

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,340 @@
...
+
+This RFC does not include support for this.
+
+### Implementation
+
+The `builtin::export_lexically` function will be used to provide the actual
+functionality. `:export` tags, if used, their options, will be used to decide
+which functions used to decide which functions are lexically exported.
+
+This module will not allow unexporting lexical functions.
```
tonycoz, line 155: Should this say "This module will not allow exporting lexical functions."?

If so, why not?  The main issue is it doesn't work with `@EXPORT_OK`, since that only includes names, but I can see it being useful to export a sub that you don't want to be a method.

Also, should that be "RFC" instead of "module"?

Ovid: @tonycoz Thank you. I've fixed the wording and pushed.

rfcs/rfcOVID-0019.md
```diff
@@ -0,0 +1,342 @@
...
+
+And more, including many modules which manually implement exporting directly
+in the `import` method.
+
+There are also many attribute-based exporters:
+
+* [Exporter::Simple](https://metacpan.org/pod/Exporter::Simple)
+* [Perl6::Export](https://metacpan.org/pod/Perl6::Export) (not really attributes)
+* [Perl6::Export::Attrs](https://metacpan.org/pod/Perl6::Export::Attrs)
+* [Exporter::Attributes](https://metacpan.org/pod/Exporter::Attributes)
```
rwp0, line 40: This is my favorite, supports both variables and functions, and also optional exporting albeit with capital `:Export` and `:Exportable` attributes.

PR #16, opened 2022-04-03
RFC 0018: eliminate need for magic true value in modules

Ovid:
This was mistakenly opened on `master` instead of a branch. This is against a branch to allow me to create more RFCs in the future without pulling in this one.

Original discussion at https://github.com/Perl/RFCs/pull/14

Comments on that RFC have been addressed in this PR.


rfcs/rfc0018.md
```diff
@@ -0,0 +1,105 @@
+# A feature to automatically "yield true" at the end of a file
+
+## Preamble
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      0018
+    Status:  Draft
```
johannessen, line 8: I suggest to follow [the proposed self-assigned ID scheme](https://github.com/Perl/RFCs/blob/main/docs/future.md#rfc-ids-and-how-to-self-assign) as long as this remains a draft: `OVID-0018`

Ovid: Thank you. I hadn't seen that doc.

How does that translate to the filename? Does it matter?

Ovid:
Per [johannessen](https://github.com/johannessen)'s comments, this has been renamed to [rfcs/rfcOVID-0018.md](https://github.com/Perl/RFCs/pull/16/files#diff-ebaaf51c679267eccd967fc2f6c0bd9aa21405f2b674b8318757557519466688)

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,105 @@
...
+
+    use feature 'yield_true';
+
+Code using the above does not need to return a magic true value when compiled.
+
+If the module explicitly returns a false value, we should consider if this can
+be ignored (I suspect this is usually a bug). However, the module author
+_might_ have conditional logic to explicitly return a false value in case
+there's a missing dependency (or for some other reason). Discussion might be
+useful here.
```
haarg, line 39: If the module author wants the module to fail to load, the proper thing to do is die with an actual error message.

rfcs/rfc0018.md
johannessen: Good question. In this particular case, I expect the RFC will leave “Draft“ status and gain an official ID quickly, so the draft file name may not matter at all.

In general, however, continuing `rfc` with no delimiter seems less than ideal, particularly when the author chooses to use a (lowercase) GitHub ID. I think inserting a hyphen would make sense (`rfc-OVID-0018.md` / `rfc-ovid-0018.md`).

rfcs/rfcOVID-0018.md
Ovid: I've fixed that in the RFC.

leonerd:
> Point 1: "yield true" is a really bad choice of name. It gives no hint at all about the fact that it's associated with the return value of require. It sounds more like something that might lexically alter the behaviour of boolean operators, perhaps? Maybe instead something like "require_returns_true"? I presume the idea is that, mostly, people won't actually be typing this; it will be enabled automatically by 'use v5.38' or whatever?

Yeah; it's unexpected that many people will actually use the feature by its name. The point is mostly just to put it into a feature bundle for a `use VERSION` declaration. And also, the name was a quick response from me to one of Ovid's earlier RFC ideas ("module"). It's likely there are better naming ideas around.

> Point 2: will it be truly lexically scoped? So if it appears in the top-most scope of a required file, it does it's thing, while in principle allowing for reversion to the old behaviour in an inner scope with a 'no feature "require_returns_true"', say?

It would be file-scoped, in practice. The only effect it has is to make a `require` of this file return true, so while you could toggle it within a smaller lexical scope of that file, it wouldn't do anything. The implementation would likely just test the state of the feature when it falls off the end of the file at the end, so any intermediate value changes it had in the middle wouldn't matter.

iabyn:
On Mon, Apr 04, 2022 at 03:59:20AM -0700, Paul Evans wrote:
> It would be file-scoped, in practice. The only effect it has is to make
> a `require` of this file return true, so while you could toggle it
> within a smaller lexical scope of that file, it wouldn't do anything.
> The implementation would likely just test the state of the feature when
> it falls off the end of the file at the end, so any intermediate value
> changes it had in the middle wouldn't matter.

But what about, e.g.

    package Foo;
    use feature 'require_returns_true';
    ...
    if (...) {
        no feature 'require_returns_true';
        return;     # ... or:
        return 0;   # ... or:
        return 1;
    }
    return;     # ... or:
    return 0;   # ... or:
    return 1;   # ... or:
    0;          # ... or:
    1;          #

Even if some of these permutations aren't useful, we still need to be
clear what happens in each case, document them, test them, and keep
them unchanged.

-- 
"But Sidley Park is already a picture, and a most amiable picture too.
The slopes are green and gentle. The trees are companionably grouped at
intervals that show them to advantage. The rill is a serpentine ribbon
unwound from the lake peaceably contained by meadows on which the right
amount of sheep are tastefully arranged." -- Lady Croom, "Arcadia"

haarg:
`require_returns_true` isn't really appropriate. `require` already always returns true.

The current RFC states that an explicit return would override this feature, meaning the feature only impacts end-of-file. So enabling or disabling the feature in an inner block would have no impact.

There are two ways this feature could work based on its current description, which would have an impact on things like naming.

The first is "current file ignores implicit return value and returns true". This is the simpler model and matches closer with the current behavior. It's essentially just a direct replacement for the `1;` at the end of a file, changing nothing else. However, it has an impact on `do` that may not be desired. Files loaded with `do` often have an implicit return value that the caller wants to use. And that would be ignored for files with a `use v5.40;` or similar which we want to be able to encourage universally.

The second option is "require of this file should ignore an implicit return and treat the return as true". This is more complex and needs more interaction between the file and the code loading it. In this model, the file is still returning the same value it always has, but if it is an implicit return, and `require` is the op checking the value, it needs to ignore that value and treat it as having returned true. And `do` would continue to return an implicit return value.

There is another model that could be used. `require` could always ignore the return value from the file if the feature was enabled. This is simpler than the previous option, but accomplishes essentially the same thing. It still needs special handling in `require`, but doesn't need to care about an implicit vs explicit return. In practice, the return value from a `require`ed file is not usable for anything. The only impact it will have on perl's behavior is throwing an error for a false value. This is better done by throwing a real error. If there is no real purpose for returning an explicit value, why complicate the model by trying to handle specially?

haarg:
As stated on #14, the PR this replaced, there is a difference between returning a false value and throwing an error. If an error is thrown, future attempts to require the same file will fail immediately without attempting to load the file at all. If a false value is returned, require will still throw an error, but attempting the load the module again will retry compiling the module.

I'm not particularly convinced that allowing the module to be recompiled is a useful feature, so I don't think the "return false" behavior needs to be facilitated under this feature. But others may disagree, and this would argue against the third option I proposed above.

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+
+Eliminate the need for a true value at the end of a Perl file.
+
+## Rationale
+
+There's no need to have a true value be hard-coded in our files that we
+`use`. Further, newer programmers can get confused because sometimes code
+_doesn't_ end with a true value but nonetheless compiles just fine because
+_something_ in the code returned a true value and the code compiles as a
+side-effect.
```
oodler577, line 27: It is my understanding that `use`, `do`, and `require` actually execute code. Sure there is the "compile" phase accessible via blocks like `BEGIN` but modules are not _loaded_ in some `perl -c` kind of way as far as I undertand it. So I think this might affect the whole point of this - a return code of some kind is actually required _or_ whatever is loading the code should be sandboxed in some kind of `eval` in order to keep the calling program from `die`'ing if things go sideways.

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+## Preamble
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      OVID-0018
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes a `yield_true` feature. When used, the current Perl _file_
```
leonerd, line 12: I'd avoid naming the feature so early; as we'll see later on the name has issues. Perhaps just put

    This RFC proposes a feature which, when used, causes the current Perl file to ...

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+
+There's no need to have a true value be hard-coded in our files that we
+`use`. Further, newer programmers can get confused because sometimes code
+_doesn't_ end with a true value but nonetheless compiles just fine because
+_something_ in the code returned a true value and the code compiles as a
+side-effect.
+
+## Specification
+
+    use feature 'yield_true';
```
leonerd, line 31: There have been a variety of comments that this name is not a good choice. Some other ideas inspired by the mailing list:

 * module_true
 * module_is_true

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+
+Code using the above does not need to return a magic true value when compiled.
```
leonerd, line 33: I'd avoid the word "magic" - that has a specific meaning in Perl which doesn't apply here.

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+
+If the module explicitly returns a false value, module loading will fail as it
```
leonerd, line 35: Do you mean "returns" as in "by use of the `return` keyword"? Might be best to spell that out specifically

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+does now. If the module author wants the module to fail to load under certain
+conditions, they should die with an appropriate error message rather than
```
leonerd, line 37: I'd suggest a slight wording change, to avoid suggesting that the author *themselves* should die ;)

Maybe something to the effect that "If the module author wants ..., they should use the `die` keyword or similar to arrange an appropriate exception to be thrown ..."

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
...
+returning false.
+
+## Backwards Compatibility
+
+There are no compatibility concerns I'm aware of.
```
leonerd, line 42: Usually best to spell out that this is because we're only suggesting changing behaviour in the presence of a newly-added feature, which is not present in any existing code.

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,104 @@
+# A feature to automatically "yield true" at the end of a file
+
+## Preamble
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      OVID-0018
```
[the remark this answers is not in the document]
johannessen: @leonerd When the PSC formally assigns IDs, could we get them listed under ["Exploratory" in README.md](https://github.com/Perl/RFCs/blob/main/README.md#exploratory), please?

Ovid:
@leonerd I believe I've addressed all of the issues you've raised. Thank you. Let me know if I've missed something. I've only been casually following the P5P discussion, so maybe I've missed something.

abraxxa (changes requested):

rfcs/rfcOVID-0018.md
```diff
@@ -0,0 +1,108 @@
...
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      0018
+    Status:  Draft
+
+## Abstract
+
+This RFC proposes a feature which, when used, causes the current Perl file to
+to yield a true value when C<require>d, eliminating the need for a "1" (or
+other true value) at the end of the file.
```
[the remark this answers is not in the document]
Grinnz: No; this is a proposal that affects module compilation, not `require`. The primary usability distinction being where the feature needs to be used to be effective.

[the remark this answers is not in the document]
abraxxa: That's not what's discussed in the 'Pre-RFC: yield true feature' p5p mail thread.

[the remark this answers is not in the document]
Grinnz: That's an extremely long thread, and I can personally attest that it has been discussed there.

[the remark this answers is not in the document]
abraxxa: Ah, it seems I haven't read Dave's proposal correctly. Switching between p5p and here is bad. 👎🏻 
I guess as it's an RfC now, continuing the discussing here is better.

Why would you remove the possibility to get the return value? Are there no use-cases for returning something from require? What about a module that is a singleton where loading returns it instead of having a constructor?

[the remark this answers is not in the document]
Grinnz: Actually it's best to keep discussion about the design on p5p. And these questions have been discussed there (though as it's an extremely long thread, I understand missing them). The return value from require is largely useless because it only returns the value if the module has not been required before in that interpreter. Also, since this is an opt-in feature you can simply not use it. If you want to be able to use a return value from loaded code, you would need to use `do`.

tonycoz:
Maybe `no feature 'require_requires_truth';`, negating it like some of the other mis-features.

rfcs/rfcOVID-0018.md
haarg: I believe this should be removed or changed. An explicit return value should be ignored, even if it false, and `require` should return a standard boolean true.

rjbs:
With apologies, I am closing this in favor of #27 because this PR did not have "allow updates by maintainers", so I couldn't make updates.

PR #19, opened 2022-06-18
RFC 0019: template literals

rjbs:


haarg:
It's implied that `$scalar` and `@array` interpolations would not happen in the `qt` string, but probably should be stated explicitly.

Also probably worth being explicit about how escape sequences work. Do all the normal backslash escapes work? And can the `{` be escaped to prevent it from being a template section?

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
+# A short and accurate title
+
+## Preamble
+
+    Author:  Ricardo Signes <rjbs@semiotic.systems>
+    Sponsor: RJBS
+    ID:      RFC 0018
```
leonerd, line 7: ITYM "0019"

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
+# A short and accurate title
```
leonerd, line 1: How about "String Literals with Embedded Code" or somesuch?

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
...
+This:
+
+```
+qt{A { TEXT } B};
+```
+
+will be equivalent to
+
+```
+qq{A ${ \scalar TEXT } B};
```
leonerd, line 68: I get where you're going with this, but a strict pedant would point out that there might be operator precedence problems here. Strictly speaking it should be equivalent to

    qq{A ${ \scalar ( TEXT ) } B};

because `TEXT` might contain something like `C() or D()`, which would parse lower than the `scalar` operator.

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
...
+
+This proposal uses `{...}` (like JavaScript) instead of `#{...}` (like Ruby)
+because the author didn't think the extra character was likely valuable enough
+for disambiguation between literal `{` and start of interpolated expression.
+
+This proposal does not offer a means to interpolate a list without (say)
+`join`.  It would require more syntax and leave the user falling back on
+remembering that multiple arguments to `print` are joined with one thing (`$,`,
+usually an empty string) but multiple list elements interpolated into a string
+are joined with another (`$"`, usually a space).
```
leonerd, line 160: FYI I think this is no bad thing. The weird inconsistency between `print @arr` using `$,` but `print "@arr"` using `$"` is already annoying enough and hard to remember. Making the user explicit about what they want here is fine.

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Future Scope
+
+`qt` is "qq but different".  If successful, it might suggest the usefulness of
+qt-like forms of qr, m, s, and other string-interpolating contexts.  This also
+suggests that another possible design for qt would be a pragma to change the
+behavior of qq *and other interpolating quote-like operators*.  This is more
+complex for the reader (because any given piece of code must be considered in
+terms of the enabled features), but eliminates the proliferation of QLOPs.
+
```
leonerd, line 146: Since we're considering adding this `qt` operator, and elsewhere there was vague mention of a `qa` or similar, I wonder if we want to declare as reserved any two-letter operator name beginning with `q`? E.g. by adding a warning like

```
$ perl -c "sub qz { ... }"
The identifier 'qz' is reserved for possible future expansion at -e line 1.
```

rwp0:
> Overall, LGTM give or take a couple of small comments.
> 
> One thing left unsaid is how to write a literal open-brace `{` character inside one of these quoted template strings. I suspect the answer would be `qt{This is in \{literal braces\}}` as per usual, but it would be good to spell that out in an example.

Or as qt({...}), no?

ivrntsv:
About heredocs. Maybe qt infront of curly braces should be dropped (`<<{END}` instead of `<<qt{END}`). Curly braces are kind of a distinctive feature of template strings and Perl doesn't allow other general quote constructs like q, qq and qx as heredoc modifier.

rfcs/rfc0019.md
```diff
@@ -0,0 +1,174 @@
...
+$s = "$problems";     # "" . $problems
+$s = "$colors->[0]";  # "" . $colors->[0]
+
+$s = "@$colors";      # join($", @$colors)
+$s = "$colors->@*";   # If postderef_qq on:  join($", $colors->@*)
+                      # Otherwise         :  "" . $colors . "->@*"
+
+$s = "Bring $object->{bring}";  # "Bring " . $object->{bring}
+$s = "Bring $object->bring";    # "Bring " . $object . "->bring}"
```
book, line 40: There's an extra `}` at the end of the comment, which shouldn't be there.

rfcs/rfc0019.md
garu: would it be safe to think of `qt(STRING)` as an optimized (and legible) shortcut to:

```perl
STRING =~ s:{(.*?)}:${ \scalar($1) }:eeg;
```
?

haarg:
> Or as `qt({...})`, no?

No. The delimiters for template sections are always `{ }`, no matter what the outer delimiters are. So `qt(a b {1 + 2} c)` and `qt{a b {1 + 2} c}` would act the same.

rfcs/rfc0019.md
rjbs: @garu "Kinda, yes" I think.  I am not going to try to think very hard about edge cases.  Roughly that.

PR #24, opened 2022-08-05
RFC process v2, draft 2

rjbs:
For more information, see [the discussion on p5p](http://nntp.perl.org/group/perl.perl5.porters/264412).


docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+## The Process
 
+![a flowchart of the process described below](/images/flowchart.png)
+
+
+### Pre-RFC
+
+The RFC process starts with an formal proposal to add or change a language feature in Perl.  But you, the prospective author of an RFC, shouldn't start by writing one.  Start by posting to p5p that you have an idea.  Explain what problem you're solving, how you think you can solve it, and what prior art you looked at.  Be clear and concise.  Make it easy for the rest of the list to see what you're suggesting without reading an enormous wall of text, but don't lose so much detail as to be meaningless.
+
+You are taking the temperature of the list.  If there is a great outcry that this is a bad idea, or has been tried before, or was explicitly rejected before, you should probably stop.  If everyone says, "Sounds great, flesh it out!", or if you're not sure, you should post, requesting that the PSC approve producing a draft.  If they do, move on to "Draft Proposal" below.  If not, they'll provide more feedback like "this is definitely impossible" or "you need to provide more information" or so on.
```
skington, line 54: The "If there is a great outcry" sentence led me to think that this paragraph would be in three parts:
* If there's a great outcry, give up
* If there's great acclaim, proceed
* Otherwise, proceed with caution

But that's not what you're saying; you're saying that if everybody says "this is terrible", give up, and otherwise move on to the next stage. So I think the word "otherwise" should be involved here; or you should split this paragraph into two, e.g.

> You are taking the temperature of the list. If there is a great outcry that this is a bad idea, or has been tried before, or was explicitly rejected before, you should probably stop. No hard feelings!
>
> If your idea isn't considered to be obviously bad, you're ready to move on to the next step. You should post a follow-up, requesting that the PSC approve producing a draft. *continue as before*

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+During this "Pre-RFC" phase, your proposal isn't in the RFC tracker.  It's not an RFC yet!
+
+During this phase, the proposer is responsible for moving things forward.  If
+they contact the PSC for approval to file a draft, and the PSC does not
+respond, it's the proposer who should be keeping track of that.
```
skington, line 60: Should this read e.g. "During this phase, you, as the proposer, are responsible"? Other parts are written in the second person.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+### Draft Proposal
+
+The PSC has agreed that you should write a formal draft proposal.  You get the template document and fill it out.  You take its advice, thinking hard about what goes in each section.  Then you post it to p5p as an email with the subject "PROPOSAL:  my great idea".  Members of the list will reply with more questions and suggested amendments.  You should read them and amend the proposal to clarify your ideas or react to valid criticism.
+
+During this phase, you (the proposer) are responsible for moving things
+forward.
+
+When you think your idea has been sufficiently scrutinized, and you gotten all the feedback you're going to benefit from, submit your updated document to the PSC.  They will either request more discussion, reject the proposal, or enter it into the tracker with the status **Exploratory**.
```
skington, line 69: Missing word: "you have gotten all the feedback".

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+### Exploratory Status
+
+When a formal draft has been discussed sufficiently, submitted to the PSC, and is under consideration, it is entered into the proposal tracker, with the status **Exploratory**.  The PSC (or other deputized folk) will begin the process of vetting the idea.  They will review discussion about the idea, they will produce a list of questions not yet answered, and they will press existing core team members (or other experts) for input.
+
+The PSC (or deputies) will eventually either:
+ * move the document to **Implementing** status (see below) because they believe it is ready for implementation
+ * move the document to **Rejected** status because they believe it should not be implemented; This may come with advice on how to formulate an alternate proposal that has more chance of being accepted.  This isn't used for "it needs some edits", but for "it's fundamentally deeply flawed."
```
skington, line 77: Ridiculously minor: I think the semi-colon before "This may come with advice" should be a full stop.

skington (commented):
Commenting purely on the new changes, and mostly about grammar; I don't know the previous document well enough to talk about substantive changes.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+ * move the document to the **Expired** status because the original proposer (or other vital personnel) are not responsive
+
+During this phase, the PSC is responsible for moving the proposal forward.
```
garu, line 80: I understand this is volunteer work (and I truly appreciate it), I just wonder if there is any sort of timeline/timeframe to be considered on the process itself (e.g. "RFCs in the exploratory status _should_ be discussed by the PSC at most in 3 subsequent PSC meetings, after which either a new status will be set, or the PSC will come forward with a list of questions. If said questions are not satisfactorily answered by the next 3 subsequent PSC meetings, it will move to either rejected or expired")

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+### Exploratory Status
+
+When a formal draft has been discussed sufficiently, submitted to the PSC, and is under consideration, it is entered into the proposal tracker, with the status **Exploratory**.  The PSC (or other deputized folk) will begin the process of vetting the idea.  They will review discussion about the idea, they will produce a list of questions not yet answered, and they will press existing core team members (or other experts) for input.
+
+The PSC (or deputies) will eventually either:
+ * move the document to **Implementing** status (see below) because they believe it is ready for implementation
+ * move the document to **Rejected** status because they believe it should not be implemented; This may come with advice on how to formulate an alternate proposal that has more chance of being accepted.  This isn't used for "it needs some edits", but for "it's fundamentally deeply flawed."
+ * move the document to the **Expired** status because the original proposer (or other vital personnel) are not responsive
```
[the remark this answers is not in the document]
garu: Maybe "expired" can be changed to "stalled" (if that's what it means? "expired" sounds to me more like rejected than "waiting for someone with enough core points to champion it into another state" - assuming that's what it means (again, sorry if my understanding was mistaken here).

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+During this phase, the PSC is responsible for moving the proposal forward.
+
+### Implementing
+
+When a proposal is accepted, the PSC is stating that they'd like to see the idea implemented, and that they are likely to merge an implementation if it doesn't bring with it any unwelcome surprises or complications.
```
garu, line 84: Who gets to implement it? Can any core developer champion it, or should it be someone assigned by the PSC? Does a main point of contact implementation-wise needs to exist before it moves to "implementing". Can they hand it to others in case they get stuck and/or lose interest? During the implementation phase, should any issue or question arise that hasn't been foreseen in the earlier stages, who calls the shots? The proposer, the implementer, both, neither of them? does it go back to an exploratory stage?

garu: Also, as stated earlier in this document, _"most work on the parser and tokeniser came from just 5 people. All of them are busy, and none of them are employed by 'Perl 5 Porters'."_ Maybe getting proposers (and other interested parties) outside that circle a chance to help out could lift some of that weight, maybe even help motivate said implementors.

I understand mentoring can backfire and add another weight to the process, but it shouldn't block other people from helping attack low hanging fruits such as writing tests, documentation, or just getting to know more about how the internals work - if only to truly grasp and appreciate the incredible work the implementors are doing, and maybe in the future become implementors themselves.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+During this phase, the proposer or other named implementor is responsible for moving the proposal forward.  The PSC will only review proposals in this status when work is delivered or when status updates cease.  The PSC will make a regular note of proposals in this status to perl5-porters.  The implementors of proposals in this status will post regular updates -- regular enough, at any rate, to avoid becoming Expired.
+
+### Testing
+
+Now there's an accepted proposal and an implementation.  At this point, it needs to have enough tests, showing how it will work in real world conditions.  The PSC will consult with the author, each other, perl5-porters, and other experts to produce a list of needed test cases.  The proposer, implementer, or other volunteers will provide the testing.  Once enough testing is done, then the document will be either…
+
+ * marked **Accepted**, with the code merged as a new (probably experimental!) feature of perl
+ * marked **Rejected**, because the quality or behavior of the feature or its implementation are not acceptable
+
+Also, of course, the document may move to **Expired** if workers vanish.
```
garu, line 101: Again, I think this needs an explicit time frame. For example, as with the "implementing" state, 3 months of inactivity.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+If an implementation is delivered and it isn't clear that it's broken as designed, the document moves to **Testing** status.
+
+During this phase, the proposer or other named implementor is responsible for moving the proposal forward.  The PSC will only review proposals in this status when work is delivered or when status updates cease.  The PSC will make a regular note of proposals in this status to perl5-porters.  The implementors of proposals in this status will post regular updates -- regular enough, at any rate, to avoid becoming Expired.
+
+### Testing
+
+Now there's an accepted proposal and an implementation.  At this point, it needs to have enough tests, showing how it will work in real world conditions.  The PSC will consult with the author, each other, perl5-porters, and other experts to produce a list of needed test cases.  The proposer, implementer, or other volunteers will provide the testing.  Once enough testing is done, then the document will be either…
+
+ * marked **Accepted**, with the code merged as a new (probably experimental!) feature of perl
+ * marked **Rejected**, because the quality or behavior of the feature or its implementation are not acceptable
```
garu, line 99: is it possible to move an RFC that failed tests back into the "implementing" state? For example, let's say one of the tests fail on an exotic-yet-supported platform perl runs on, exposing an issue on the implementation that requires a big refactoring. Should it go back to "implementing" or stay "testing" during the fix?

garu (changes requested):

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+ * move the document to **Rejected** status because they believe it should not be implemented; This may come with advice on how to formulate an alternate proposal that has more chance of being accepted.  This isn't used for "it needs some edits", but for "it's fundamentally deeply flawed."
+ * move the document to the **Expired** status because the original proposer (or other vital personnel) are not responsive
+
+During this phase, the PSC is responsible for moving the proposal forward.
+
+### Implementing
+
+When a proposal is accepted, the PSC is stating that they'd like to see the idea implemented, and that they are likely to merge an implementation if it doesn't bring with it any unwelcome surprises or complications.
+
+If no implementation has made progress for three months, the document moves to **Expired**.
```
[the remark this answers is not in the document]
leonerd: There ought to be some sort of resurrection process, yes. Maybe via a "is this still a good idea?" test, depending how long it has been since it expired.

docs/process.md
leonerd: I expect a lot of the state-transition edges could allow backwards stepping, yes; if on closer inspection the overall idea is found to be less ready for that new state than it was first thought, then the correct thing to do is probably just to send it back again.

docs/process.md
leonerd: +1

leonerd (commented):
Overall I think this is a good description of a process we can work with. I think it might benefit a couple of notes permitting "steps backwards", and some thought about Expired vs. Stalled, and any steps required to resurrect a stalled RFC back into life again.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
+
+During this "Pre-RFC" phase, your proposal isn't in the RFC tracker.  It's not an RFC yet!
+
+During this phase, the proposer is responsible for moving things forward.  If
+they contact the PSC for approval to file a draft, and the PSC does not
+respond, it's the proposer who should be keeping track of that.
+
+### Draft Proposal
+
+The PSC has agreed that you should write a formal draft proposal.  You get the template document and fill it out.  You take its advice, thinking hard about what goes in each section.  Then you post it to p5p as an email with the subject "PROPOSAL:  my great idea".  Members of the list will reply with more questions and suggested amendments.  You should read them and amend the proposal to clarify your ideas or react to valid criticism.
```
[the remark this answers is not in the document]
rjbs: I don't think it needs to stand out more.  It's just a part of the sentence.

docs/process.md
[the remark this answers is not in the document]
rjbs: This text has changed as part of another suggestion.

docs/process.md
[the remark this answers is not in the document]
rjbs: * I have updated the text to say it's submitted by email to perl5-porters
* this document doesn't say anything about RFC numbers; I think they're stupid and don't want to enshrine them in the process
* I have updated the text to suggest feedback is made by email reply
* if the PSC isn't being timely in replies, it's the responsibility of the proposer to push

docs/process.md
rjbs: I'm not opposed, but mostly I think that a big goal for me here is to make it clear whose responsibility it is, and setting a time frame is redundant and possibly bad.  All "responsible for moving forward" should carry with it "as rapidly as is practical and responsible."

docs/process.md
rjbs: I didn't include any text because I don't think it's clear what we'd include.  The PSC *cannot assign anybody*.  That's not within our capabilities.  Suggesting it is possible would be doing everybody a disservice.  We can't just rack up great ideas, and they need to get built.  The RFC process is a way of saying "We agree this is sufficiently described to be implemented and get merged" to make implementing more attractive, but I don't think anybody has power to make implementation happen other than by *doing it*.

docs/process.md
rjbs: Done.

docs/process.md
```diff
@@ -41,105 +42,65 @@ Costs are
...
-
-Don't panic: changes can be reverted.
+## The Process
 
+![a flowchart of the process described below](/images/flowchart.png)
+
+
+### Pre-RFC
+
+The RFC process starts with an formal proposal to add or change a language feature in Perl.  But you, the prospective author of an RFC, shouldn't start by writing one.  Start by posting to p5p that you have an idea.  Explain what problem you're solving, how you think you can solve it, and what prior art you looked at.  Be clear and concise.  Make it easy for the rest of the list to see what you're suggesting without reading an enormous wall of text, but don't lose so much detail as to be meaningless.
```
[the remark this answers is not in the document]
rjbs: 1. done
2. "pre-rfc" is not a noun; it is the period of time before the document is being tracked by the PSC
3. the "one" is "a formal proposal", which I think is the only noun it could refer to; I will replace it with the noun
4. this seems overly prescriptive to me

rjbs:
I used "Expired" instead of "Stalled" because "Stalled" implies it will start again.  I believe that most things left untouched for three months won't, and acting like we're waiting for them to start again is going to create drag and a sense of half-finished things lying around, instead of things we've put behind us.  Alternatively, I would suggest "Abandoned," which still suggests somebody could come back and pick it up, but without the expectation.

docs/process.md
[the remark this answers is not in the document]
rjbs: See comment on the PR itself.

docs/process.md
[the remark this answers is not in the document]
rjbs: Done.

docs/process.md
garu: Agreed!

docs/process.md
garu: > The RFC process is a way of saying "We agree this is sufficiently described to be implemented and get merged" to make implementing more attractive, but I don't think anybody has power to make implementation happen other than by _doing it_.

I like it. If this was written in the document as is, it would be enough (for me). My understanding was that the PSC would manage the implementation stage more closely, which I why I brought it up. If, once the RFC is accepted, anyone can try and implement it, then this mention (together with some minor pointers to where interested parties should direct their implementation conversation - like irc or p5p) would be nice.

PR #27, opened 2022-09-25
RFC 0018: module true

rjbs:
This PR replaces #16, mostly because rjbs couldn't push to that one's branch to add requested updates.


haarg:
I think this should be more clear on what the return value of require should be. The name and the comments under "Rejected Ideas" seem to support that require should return "true" (PL_sv_yes). But the "Specification" does not address this, only saying that require should not not croak.

chromatic:
I've written a draft implementation in [PR 20365](https://github.com/Perl/perl5/pull/20365).

This does not address @haarg 's comment, but it's possible to do so if it's so desired.

PR #29, opened 2022-10-14
RFC 0006 - Rename `load` to `load_module`

leonerd:
Objections were raised at the rather short and inspecific `load`; suggested that we rename this to `load_module`.

### DIFF
--- a/rfcs/rfc0006.md
+++ b/rfcs/rfc0006.md
@@ -1,4 +1,4 @@
-# Module Loading with "load"
+# Module Loading with "load\_module"
 
 ## Preamble
 
@@ -9,8 +9,8 @@
 
 ## Abstract
 
-This RFC proposes a new built-in, "load", to load modules by module name
-(rather than filename) at runtime.
+This RFC proposes a new built-in, "load\_module", to load modules by module
+name (rather than filename) at runtime.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


Grinnz:
use and require have significant precedent, one might say moreso than any other part of the language. `load` is a meaningless term to Perl coders. the cases are not at all comparable or related.

leonerd:
This is about the not-yet-written `builtin::load` function, which we feel should be called `builtin::load_module`, to live with the other builtins.

rwp0:
Yes, I know. If its fitting for `builtin::` to have keywords with harder names, let it be so.

But generally having one-word keywords is better to have in core IMHO matching the other keywords of the language (it was designed like that I believe).

leonerd:
@rwp0 
> Yes, I know. If its fitting for builtin:: to have keywords with harder names, let it be so.

Who knows - maybe one day we'll add a `builtin::load_json` or whatever; at which point having named this `load_module` rather than simply `load` would feel like much of an improvement.

cpansprout:
I just want to mention that we have traditionally reserved underscores for user-defined names; perl core functions don’t use underscores.

Grinnz:
this isn't a core keyword, it's a function in the builtin namespace

rwp0:
Yeah, whatever, naming disputes shouldn't hinder the development. If the default `load` is too ambiguous, let it be `load_module`.

leonerd:
@rjbs @book - thoughts?

book:
I have no issue with the `load_module` name.

rwp0:
I then discovered this Perl API C function:

https://perldoc.perl.org/perlapi#load_module

That probably will be a front-end to this one.

Then the name is not so bad I think.

PR #32, opened 2023-01-03
Ovid/assignment context

Ovid:
RFC for adding "assignment context" attributes to safely replace `wantarray` heuristics.

### DIFF
--- a/rfcs/rfc0023-assignment-context.md
+++ b/rfcs/rfc0023-assignment-context.md
@@ -1,0 +1,179 @@
+# Assignment context attributes
+
+## Preamble
+
+    Author:  Curtis "Ovid" Poe <curtis.poe@gmail.com>
+    Sponsor:
+    ID:      0023
+    Status:  Draft
+
+## Abstract
+
+Add `:scalar`, `:list`, and `:void` attributes for subroutines/methods
+to ensure their results are handled correctly rather than relying on the
+`wantarray` heuristic.
+
+## Motivation
+
+* Create attributes for subroutines which require their return values to be
+  immediately assigned or ignored.
+* Not handling the context correctly will be a fatal error
+* Reduce dependency on fragile `wantarray` heuristics.
+* Reduce boilerplate code being hand-written over and over. 
+
+## Rationale
+
+It's very easy to write `wantarray` code that doesn't really do what you
+want. For example, I might have code which returns data which must be handled,
+so I use `wantarray` to guarantee that it's not called in void context:
+
+    sub munginator {
+        if ( !defined wantarray ) {
+            die "Do not call me in void context";
+        }
+        ...
+        return $some_scalar;
+    }
+
+The intent is that this should `die`:
+
+    my $value = get_value();
+    munginator($value); # fatal error!
+
+And this should work:
+
+    my $value = get_value();
+    $value = munginator($value);
+
+But here we have an edge case:
+
+    if ( munginator($value) ) {
+        ...
+    }
+
+That's evaluated in scalar context, so our `wantarray` heuristic has failed.
+We would like to not only stop reliance on a dodgy heuristic, but to not have
+to keep writing extra code to handle this case.
+
+Instead, it would be nice to write the following:
+
+    sub munginator :scalar {
+        ...
+        return $some_scalar;
+    }
+
+Any code calling `munginator` must do so by assigning the return value to a
+scalar. Otherwise the `if ( munginator($value) ) {...}` code fails as
+intended.
+
+## Specification
+
+Three new attributes are added, enabled via a feature named
+`assignment_context` (a better name would be good).
+
+1. `:scalar` - Return value must be assigned to a scalar, in scalar context
+2. `:list` - Return value must be assigned to a list
+3. `:void` - Return value must not be assigned to anything
+
+These attributes can be used on named subroutines/methods:
+
+    sub foo :list {...}
+
+Or anonymous subroutines:
+
+    my $foo = sub :void {...};
+
+Or a block `eval` (this is for completeness and may be be needed):
+
+    my $result = eval :scalar {...};
+
+The assignment must be in the statement calling the subroutine and cannot
+propagate. Thus, this will not work:
+
+    my $result = foo();
+    sub munginator :scalar {...}
+    sub foo {
+        return munginator(42);
+    }
+
+The above would need to be written similarly to this:
+
+    my $result = foo();
+    sub munginator :scalar {...}
+    sub foo {
+        return my $result = munginator(42);
+    }
+
+## Backwards Compatibility
+
+I am not aware of any core code which relies on these attributes. It's
+entirely possible (even likely) that there are CPAN or DarkPAN modules which
+do.
+
+## Security Implications
+
+None that I'm aware of.
+
+## Prototype Implementation
+
+None.
+
+## Open Questions
+
+What should this do?
+
+    my ($value) = some_sub();
+
+If `some_sub()` is marked as a `:list`, the above is correct, but if
+`some_sub()` returns more than one element, the behavior may not match
+developer expectations.
+
+If `some_sub()` is  marked as `:scalar`, the above should still
+work, but the assignment is in list context, not scalar context.
+
+# Exaamples
+
+    # munges arguments directly and has no guaranteed return value,
+    # so calling this is void context is required
+    sub lc_values :void ($hashref) {
+        $_ = lc $_ for values %hashref->%*;
+    }
+
+    # Returns copy of the hash
+    # so calling this is list context is required
+    sub lc_values :list (%hash) {
+        my %new_hash = map { $_ => lc $hash{$_} } keys %hash;
+        return %new_hash;
+    }
+
+    # Returns copy of the hashref
+    # so calling this is scalar context is required
+    sub lc_values :list ($hashref) {
+        return { map { $_ => lc $hashref->{$_} } keys $hashref->%* }
+    }
+
+## Prototype Implementation
+
+At the present time, creating a prototype implementation would likely require
+XS code and heavy knowledge of the Perl internals. I cannot do this.
+
+## Future Scope
+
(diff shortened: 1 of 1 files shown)


rfcs/rfc0023-assignment-context.md
```diff
@@ -0,0 +1,179 @@
...
+intended.
+
+## Specification
+
+Three new attributes are added, enabled via a feature named
+`assignment_context` (a better name would be good).
+
+1. `:scalar` - Return value must be assigned to a scalar, in scalar context
+2. `:list` - Return value must be assigned to a list
+3. `:void` - Return value must not be assigned to anything
```
leonerd, line 76: Oh. I see. That's entirely different to the idea I had in mind. What I was thinking was more that the attribute will force the context of a `return` expression or the final statement of the code body, into always being that context rather than determined at runtime based on the caller. In addition, the compiler can emit warnings (or errors), or the runtime can do so, if a function is called from a context that would be incompatible.

An "incompatible" context has to be a bit flexible though. It's perfectly fine to invoke a `:scalar` function from a list context - just remember it yields exactly one value. This is often what you want and is often fine - don't forget the list context of a function call's arguments:

```perl
sub get_thingy :scalar { ... }
SomeClass->new( thingy => get_thingy() );   ## this should be fine and allowable
```

All of these would then have nothing to do with the actual operation of assignments - that's different.

We could still as a separate idea (either in a new RFC or as part of this one) steal a function attribute from C compilers, which creates a warning if a function's result is unused; effectively, warn if invoked in void context. Sometimes it's OK - sometimes functions return an *optional* result value but it's semantically fine for the caller to ignore that, as it has done whatever work it needed to do as a side effect. At other times, it's very important that caller uses the value returned; calling the function in void context should be considered an error:

```perl
sub get_thingy :scalar :warn_unused { ... }

my $thing = get_thingy();  ## is fine
get_thingy(); 1;  ## warns (or fails to compile?)
```

tobyink: As with @leonerd, I'd want to preserve the sanctity of:

```perl
sub get_thingy :scalar { ... }
SomeClass->new( thingy => get_thingy() );
```

And also, this should work:

```perl
sub get_thingy :scalar { ... }
sub another_thingy :scalar { ... }
my $result = get_thingy() + another_thingy();
```

Ovid: OK, this seems reasonable. If I update the specification to state that `:scalar` succeeds even if called in list context, it seems like it would satisfy these concerns, but what about the propagation issue?

```perl
foo();
sub munginator :scalar {...}
sub foo {
    return munginator(42);
}
```

Since the above ultimately is in `:void` context, it still fails, but if assigned to a value at the very end, it succeeds, yes? This means that "assignment context" is a bad name for this feature.

As for @leonerd's question:

```perl
sub get_thingy :scalar :warn_unused { ... }

my $thing = get_thingy();  ## is fine
get_thingy(); 1;  ## warns (or fails to compile?)
```

I don't think we need `:warn_unused` unless we want this as a migration tool to have warnings. Then, developers can later remove the `:warn_unused` to get fatal behavior, but at their convenience. But then, we want this for the other contexts, yes?

```perl
sub get_thingy :scalar(warn) { ... }
```

bbrtj: I like what @leonerd said, but wouldn't it be better to move it one step further?
`:scalar` would require an explicit return with a single scalar, `:list` with a list. `:void` would return nothing, but bare `return;` is allowed to exit early.

I feel like this is something worth having and would likely feel odd if it involved an extra attribute (so two in total).

Grinnz: Not necessarily. A single scalar is a list in list context, and an array is a single value (its size) in scalar context. These are perfectly reasonable return values, as is an empty list (nothing) in list context.

Grinnz: There are two possible useful features here: coercion of return context, and warning/error on strictly incompatible return context. Let's not introduce anything unperlish further than that.

Ovid: @Grinnz can you provide code examples of what you mean by this?

> There are two possible useful features here: coercion of return context.

Grinnz:
```
sub get_thingy :scalar { return; }
sub get_count :scalar { return @things; }

my %things = (thingy => get_thingy(), count => get_count());
```

effectively avoiding the injection vulnerability of using functions in an unexpected context.

rfcs/rfc0023-assignment-context.md
```diff
@@ -0,0 +1,179 @@
...
+
+These attributes can be used on named subroutines/methods:
+
+    sub foo :list {...}
+
+Or anonymous subroutines:
+
+    my $foo = sub :void {...};
+
+Or a block `eval` (this is for completeness and may be be needed):
```
rwp0, line 86: spelling: `may be be needed`

rfcs/rfc0023-assignment-context.md
```diff
@@ -0,0 +1,179 @@
...
+    my ($value) = some_sub();
+
+If `some_sub()` is marked as a `:list`, the above is correct, but if
+`some_sub()` returns more than one element, the behavior may not match
+developer expectations.
+
+If `some_sub()` is  marked as `:scalar`, the above should still
+work, but the assignment is in list context, not scalar context.
+
+# Exaamples
```
rwp0, line 134: spelling: double aa in "examples"

rfcs/rfc0023-assignment-context.md
```diff
@@ -0,0 +1,179 @@
+# Assignemnt context attributes
```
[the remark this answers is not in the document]
Ovid: Agreed. I see now that it was a mistake to write it like that. Will fix.

iabyn:
I still don't understand this proposal at at all. I don't see what problems it solves. The example shows the function used in boolean context, and this is understood to be a Bad Thing, but without explaining why it's actually bad.

I think the whole idea of 'assignment context' is murky and will have endless edge cases. Does `unshift @a, foo()` count as an assignment? Etc.

tobyink:
Honestly, I think a better solution, even though it will only kick in at run-time (while Ovid's could theoretically be checked at compile time) is allowing destructors to throw real exceptions which don't get downgraded to warnings.

How does this help Ovid's use case? Say `munginator` would normally return a string. Instead, it could return an object blessed into this class:

```perl
package Local::Munginator::Result {
  use overload
    q[""]    => 'to_string',
    q[bool]  => 'to_bool',
    fallback => 1;
  sub new ( $class, $string ) {
    bless [ 0, $string ], $class;
  }
  sub to_string ( $self, @args ) {
    # @args is a dummy because overloaded methods will always be
    # called with three parameters!
    $self->[0]++;
    return $self->[1];
  }
  sub to_bool ( $self, @args ) {
    return builtin::true;
  }
  sub DESTROY ( $self ) {
    die "Munginator result was ignored!" unless $self->[0];
  }
}
```

This allows `munginator` to not just check that its result hasn't been immediately discarded, but enforce that the caller has performed certain required operations on it. (In this case, force that it's been stringified at least once.)

tonycoz:
> I still don't understand this proposal at at all. I don't see what problems it solves. The example shows the function used in boolean context, and this is understood to be a Bad Thing, but without explaining why it's actually bad.

One obvious case is DBI::connect():

```
if (DBI->connect(...)) {
    # well, I connected, now what?
}
```
For me Imager is related too:
```
# scale warns in void context to try to detect calls in void context,
# which people did.  That doesn't help here
if ($img->scale(...)) {
    # discarded the scaled image, oops
}
```

> I think the whole idea of 'assignment context' is murky and will have endless edge cases. Does unshift @a, foo() count as an assignment? Etc.

There seems to be two different concepts discussed here (which @Grinnz discusses above), the boolean check covers one side of it, forcing scalar/list/void context on return seems to me to be only indirectly connected.

Using `:scalar` to disallow boolean context seems incorrect to me - boolean context is scalar context.

I'd expect anything that actually saves the returned value to be accepted, including use as a function argument, based on the discussion above.

Also, while the above discusses boolean context as a problem - use in numeric or string context could be similar problem if the function returns a non-overloaded reference.

MartijnLievaart:
Op 12-01-2023 om 04:58 schreef Tony Cook:
>
>     I still don't understand this proposal at at all. I don't see what
>     problems it solves. The example shows the function used in boolean
>     context, and this is understood to be a Bad Thing, but without
>     explaining why it's actually bad.
>
> One obvious case is DBI::connect():
>
> |if (DBI->connect(...)) { # well, I connected, now what? } |
>
> For me Imager is related too:
>
> |# scale warns in void context to try to detect calls in void context, 
> # which people did. That doesn't help here if ($img->scale(...)) { # 
> discarded the scaled image, oops } |
>
>     I think the whole idea of 'assignment context' is murky and will
>     have endless edge cases. Does unshift @A <https://github.com/A>,
>     foo() count as an assignment? Etc.
>
> There seems to be two different concepts discussed here (which @Grinnz 
> <https://github.com/Grinnz> discusses above), the boolean check covers 
> one side of it, forcing scalar/list/void context on return seems to me 
> to be only indirectly connected.
>
> Using |:scalar| to disallow boolean context seems incorrect to me - 
> boolean context is scalar context.
>
> I'd expect anything that actually saves the returned value to be 
> accepted, including use as a function argument, based on the 
> discussion above.
>

So what is actually wanted is something like :nodiscard?


HTH,

M4

tonycoz:
> So what is actually wanted is something like :nodiscard?

That covers completely ignoring the result (void context), but not boolean/numeric/string context that might discard the useful part of a reference.

Your mention did get me to look at the C++ nodiscard again, which reminded me that can optionally take a message, so you can explain why void/boolean/numeric/string context is bad.
```
sub DBI::connect :noboolean("discarded returned database handle") {
  ...
}
```

Maybe ":nodowngrade" to cover all the cases, though downgrade is overloaded by strings, at least for me.

PR #33, opened 2023-01-18
RFC for allowing alternate iteration variable in map and grep

haarg:


demerphq:
Nice. This is an RFC that really should happen.

haarg:
There is one thing I put in "Future Scope" which some could consider a larger problem - not being able to use this syntax with `(&@)` prototyped subs. I left it as something for the future rather than a pressing issue because I feel like the syntax proposed by this RFC is a natural extension how `for` works. `(&$)` is a bit of an imperfect hack, and trying to fit it with this concept will likely require something else that is an imperfect hack.

But some may consider that instead, the `{ ... }` section of a map/grep should be considered more as akin to a sub, and being able to specify a signature for it. And then extending signatures to allow checking their arity. This concept would push more for a compact inline sub syntax, similar to javascript's arrow functions.

haarg:
Syntactically, a prototype is parsed like a `q()`, which means you can put basically anything in it, including nested parenthesis.
```
$ perl -E'say prototype sub (lol (welp)) { }'
lol (welp)
```
Defining a function with an invalid prototype will produce a warning if they are enabled. Trying to call a function with an invalid prototype will produce a parse error, if the function call is bound early.

demerphq:
So couldn't we come up with a symbol to handle this so that folks could write their own?

haarg:
Yes, it should be possible to create some new kind of prototype that allowed this syntax. Deciding how that would actually work and be passed into the function could then be left as "Future Scope".

demerphq:
@haarg: right. my point being that this proposal doesn't slam any doors shut, it just means we need to some work before the door can be opened.

haarg:
The only reason it could give pause is if it was indicating that the concept of this RFC was wrong. This RFC proposes "make map more like for", where someone else might think we should "make map { ... } more like a sub call".

demerphq:
@haarg are the two concepts necessarily mutually exclusive? Eg, we could make the "m" prototype magic in the sense that it would Just Work with both forms.

wchristian:
Loving this. :)

The examples section probably should also show some non-block form examples so readers will ponder these as well. (Also there's an errant lone space line in the examples section.)


```perl
my @my_array = (11 .. 14);
my @trad_map = map $_ + 1, @my_array;
# ( 12, 13, 14, 15 )
 
my @new_map = map my $f $f + 1, @my_array;
# ( 12, 13, 14, 15 )

my @pair_map = map my ($f, $g) $f + $g, @my_array;
# ( 23, 27 )

my @new_grep = grep my $f $f > 12, @my_array;
# ( 13, 14 )

my %my_hash = (
   first_key => 11,
   second_key => 14,
);
my %pair_grep = grep my ($f, $g) $f =~ /sec/, %my_hash;
# ( second_key => 14 )
```

haarg:
I don't think it makes sense to try to support this syntax with the expression form of map or grep. I could make that more clear in the RFC I suppose.

wchristian:
As per discussion on IRC:

I maintain PPI and i am looking at this from a perspective of how Perl code is read and parsed on a conceptual level, and also as someone who actually uses the expression map form, and for technical reasons and not just aesthetic.

The core proposal of the RFC is strictly only:

> allow inserting an extra term directly after the word map, to modify behavior,
> then parse the rest identically as before

This means that semantically this:
```perl
my @trad_map = map { $_ + 1 } @my_array;
my @trad_map = map $_ + 1, @my_array;
```

Is naturally twinned by:
```perl
my @trad_map = map my $f { $f + 1 } @my_array;
my @trad_map = map my $f $f + 1, @my_array;
```

The actual code implementation might have a bit more trouble with that, but this is what happens on a language design level. Your position is currently that `map my $f $f + 1, @my_array` should not work at all. Semantically this would be a special case, an exception.

This comes with two problems:

1. This would be surprising to users. People will want to use this, and expect it to work. I am patient zero for this, but there will be many more. Semantically there is no reason to expect this to fail and such a failure can only be a surprise.
2. This would gate the named form from people who use the debugger in certain ways. The way the Perl debugger works is that when iterating over a map, there are two commands to step forward. One steps into the current statement, one steps over the current statement to the next one. The latter, `n`ext, pauses on every iteration when invoked on `map { $_ + 1 } @list`, but steps over all iterations when invoked on `map $_ + 1, @list`. Forbidding naming as in `map my $f $f + 1, @list` would result in a developer who uses the debugger needing to decide whether to be able to step over map calls, or use naming; often resulting in not being able to use naming at all.

Your only technical argument against it so far was "implementing the perl parser bits would be a bit more complicated". The only other argument you gave me, as per your own words, paraphrased: "I am against it because of my personal preference, and you will not convince me." I see how implementation is a cost, and i could be convinced that it is just much too complicated for the perl core to even try to implement the semantical twinning. However if the cost is not prohibitive, I don't see how that balances out the harm described above.

Aesthetically there were some arguments against it, to which the replies are:

1. `map my $f $f + 1, @list` is not how it would be used mostly. In single-variable naming contexts, $_ is fine, or the expression would be more complicated, like:
```perl
my @prepared = map my $step $self->curry::run_sub_step( $step, $request, $eventloop, $uid ), $groups->@*;
```
2. For the most likely use-case, multi-variable-naming, the aesthetics are fine as there is visual separation:
```perl
my @pair_sums = map my ( $f, $g ) $f + $g, @list;
```
3. Users of map who actually use the term form already do often decide to visually separate on their own via `map +( $_ => 1 ), @array`, and given that Perl's philosophy is to give people plenty of rope and trust them to be smart with it, i don't see why in this case we would decided to actively distrust them and not expect them to use `map my ( $f, $g ) ( $f + $g ), @list;` in cases where the code would look ugly.
4. @leonerd brought up the complaint that map already has ambiguity, specifically around the question of "is `{}` a hashref or a block?" However this does not actually impact the decision of whether to allow naming on expression-form map, since that problem already exists, and forces people to use `map +{ $_ => 1 }, @list` or `map( { $_ => 1 }, @list )` just to ensure the perl parser doesn't bring them trouble. As such it is nothing new and not something that actually interacts with allowing naming on expression-form map.
That said i do agree that that is troublesome and in future under `use v7;` after `map ` or `map my $f ` any `{`  should be seen as the start of a block, and people who wish it to be a hashref would need to start with `map my $f +{` or `map my $f ( {`. I plan to make an RFC for that, but this should ameliorate concerns around that, even if as is it is only relevant tangentially.

All of that said: These arguments for forbidding naming with the expression form are neither technical nor even language design related, they are, as agreed to me, personal preferences on how perl code should be written. I do not think that should guide language design.

In the end though, i am happy to be convinced. If the technical cost is just too much or there are things i am as of yet unaware of, it might be the right choice to forbid.

haarg:
I think syntax like `map my $f $f + 1, @list` is too ugly to allow, and doesn't fit any established language conventions. It changes expression form map from using (basically) normal function call parsing to instead be something else with no precedent.

If there are problems using the debugger on block form maps, that sounds like something that should be addressed in the debugger.

I have no intention of extending this RFC to include the expression form of map/grep.

leonerd:
> I think syntax like `map my $f $f + 1, @list` is too ugly to allow
>...
> I have no intention of extending this RFC to include the expression form of map/grep.

+1 to everything @haarg just said.

leonerd:
**If** we could have some new kind of arrow syntax that is distinct from anything else Perl currently uses, I could imagine a "lightweight lambda" arrow that we could use for these. E.g. here's a random one I borrowed from a Unicode chart:

```
my @squares = map my $x ⇨ $x ** 2, @numbers;
```

then maybe *maybe* in that situation we might consider lightweight expression statements like this.

But we're stretched far to thin on the limited supply of symbols that ASCII gives us, and apparently folks would complain if we started including syntax that *requires* non-ASCII Unicode. So that idea is basically dead in the water.

wchristian:
> I think syntax like `map my $f $f + 1, @list` is too ugly to allow,

I have no idea why any person doing programming language design would say something like that and i do not think that is nor should be how Perl language design happens.

It also completely ignores what i said in point 1.

> and doesn't fit any established language conventions. It changes expression form map from using (basically) normal function call parsing to instead be something else with no precedent.

Expression form map is not and never was a normal function call. It may take the shape of one, but semantically it is an implicit wrapping of delayed code that in itself has no simile anywhere else in Perl.

> If there are problems using the debugger on block form maps, that sounds like something that should be addressed in the debugger.

Being that that inherently relies on the difference between the block and the implicit wrapped delayed code, i do not think that can be fixed.

> I have no intention of extending this RFC to include the expression form of map/grep.

I remain convincible, but barring actual technical arguments all i have gotten so far is "i don't like it so nobody else should have it", which is incredibly selfish and willfully harmful and will continue to express this at all instances.

wchristian:
@leonerd
> **If** we could have some new kind of arrow syntax that is distinct from anything else Perl currently uses, I could imagine a "lightweight lambda" arrow that we could use for these. E.g. here's a random one I borrowed from a Unicode chart:
> 
> ```
> my @squares = map my $x ⇨ $x ** 2, @numbers;
> ```

If the only complaint is visual separation then there is no need to have it be some kind of symbol. What's good for python
```py
for x in "banana":
  print(x)
```

can also be good for perl
```perl
my @squares = map my $x in $x ** 2, @numbers;
```

> then maybe _maybe_ in that situation we might consider lightweight expression statements like this.

You don't have to consider adding it, you'd have to consider expressly forbidding it and making the syntax inconsistent and then justify that.

wchristian:
Thanks Yves. :)

I'm not a native speaker, and have no strong preference for what kind of separator it should be and am happy to accept anything that others think feels better. A small thought is that i would prefer one that is easy to type on a german keyboard, but that is a very small one.

wchristian:
One more question:

Is this also objectionable? If so why?

```perl
    my @prep = map my ( $step, $uid ) $self->curry::run_sub_step( $step, $request, $eventloop, $uid ), $groups->@*;
```

Grinnz:
In my opinion the expression form in either case does not flow well without a separator, but with a separator (arrow or word) is acceptable.

demerphq:
On Fri, 20 Jan 2023, 03:36 Christian Walde (Mithaldu), <
***@***.***> wrote:

> One more question:
>
> Is this also objectionable? If so why?
>
>     my @prep = map my ( $step, $uid ) $self->curry::run_sub_step( $step, $request, $eventloop, $uid ), $groups->@*;
>
>
I haven't tested, but I think it's likely almost valid syntax and it would
lead to parse conflicts. The need for an additional operator after the my
allows us to disambiguate. map is already problematic as it can't
disambiguate a block from a hash constructor until it hits the comma. I
have some sympathy with the argument we shouldn't make this worse, but
leonerds point that some kind of arrow or yours of adding a keyword after
the my expression likely would resolve the conflict. Grammar stuff is hard
though and I may be missing something. Leonerd would know better I imagine,
or zefram would if you could get him to answer.

Yves

>

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,180 @@
...
+
+### `map my (VAR, VAR) BLOCK LIST`
+
+This will evaluate `BLOCK` for each set of two elements in `LIST`, aliasing
+the `VAR` to the first of each set, and the second `VAR` to the second. If
+there is not a multiple of two items in `LIST`, the last run of `BLOCK` will
+assign `undef` to the extra `VAR` slot not corresponding to an element in
+`LIST`. More than two variables can be used to iterate over sets of three or
+more items.
+
```
leonerd, line 84: To answer the questions from the comment above, it might be helpful to spell out the behaviour in more detail about `grep my VAR BLOCK LIST` and `grep my (VAR, VAR) BLOCK LIST` here.

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,180 @@
...
+```perl
+my @out = map my ($key, $val) {
+  map my $i { "$key: $i" } $val->@*;
+} %hash;
+```
+
+All of this would apply to `grep` as well.
+
+As the syntax proposed is currently invalid, a feature should not be needed to
+enable it.
```
leonerd, line 59: As much as I'd love to see this syntax, I have my slight doubts here. Under current 5.36:

```
eval: my @result = map my $x = $_*2, (1,2,3);  \@result
[ '2', '4', '6' ]
```

meaning that the parser is going to have to at least lookahead past the `my` and the `$x` in order to find the `{` or lack thereof. Given that you want to permit an arbitrary sized list of scalars (and maybe `undef` as well) this is a potentially long amount of lookahead. There may be further pathological cases I've not thought of.

Even if it is possible to implement it by some scary amount of code in perl's tokenizer, I'm not convinced we'd *want* to do that even so. I have a particular bee in my bonnet at the moment about trying to make Perl syntax more predictable and parsable with [simple(r) parsers] so I might be biased here... but...

I can imagine that it might require a named feature to enable this syntax while simultaneously disabling the use of `map EXPR, LIST`. I'd be happy to consider such an option. For all the new abilities such a syntax gives (as you point out, it entirely replaces `pairmap` and `pairgrep` for example), I'd be happy to have to pay for that by requesting it.

```perl
use feature 'extended_map';

my @squares = map my $x { $x **2 } @numbers;
```

I believe the RFC should at least suggest a name for a feature guard, if it turns out that one is required, so it can be discussed ahead of time rather than ad-hoc during development.

[simple(r) parsers]: https://github.com/tree-sitter-perl/tree-sitter-perl/

leonerd (commented):
We discussed this in the PSC meeting today. Here is a summary of my thoughts.

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,180 @@
...
+A natural extension of this syntax would be to allow n-at-a-time iteration, as
+`for` can do on perl 5.36+.
+
+```perl
+my @out = map my ($key, $val) {
+  map my $i { "$key: $i" } $val->@*;
+} %hash;
+```
+
+All of this would apply to `grep` as well.
```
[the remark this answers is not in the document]
Grinnz: I agree with both feelings. I would consider the "undef" placeholders to be more of the result of needing something where there's nothing, rather than being treated as actual undef values; thus they should have no effect on the ultimate result when reasonable.

hvds:
Paul Evans ***@***.***> wrote:
> This does raise an interesting cornercase question: What does n-at-a-time `grep` mean in scalar context? Does the result count the total number of values it would have returned, or the number of *bundles*? Example:
> 
> ```perl
> my $count = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
> ```
> 
> Does `$count` now equal 1 or 2?
> 
> My gut-reaction feeling is that it should yield 2 in this case, thus retaining the invariant that assigning the result to a temporary array and then counting the array is another way to give you the same count.

I'm not sure that invariant is something useful to preserve in this case.

My expectation for scalar mode would be for it to return the number of times
that the block yielded a true value. I tend to want to know "for how many
things is X true", where a "thing" is whatever I'm applying X to - in this
case X is the block expression, and the "thing" is a bundle.

I don't really imagine I'd ever want to count the number of bundled elements,
but if I needed it I'd multiply by n in the well-formed case and use
something like `my $count = () = grep ...` for the trailing-undefs case.

wchristian:
@demerphq thanks for pointing that out. tho from what leonerd is saying about the parser being unable to handle long constructs that aren't preceded by unambiguous differentiators points towards a trailing arrow or simile not being useful in this case

i had other thoughts here, but that got derailed to hell by this suggestion:

> it might require a named feature to enable this syntax while simultaneously disabling the use of map EXPR, LIST

the downstream consequences of that would be very destructive. proposing that as a first line consideration instead of trying or even asking for compromises feels cartoonishly evil and selfish to me and is frankly making me want to give up if this is considered to be an acceptable way to treat users of perl. you can do better

that said, in the spirit of actually trying to be constructive:

what about

```perl
map :my( $f, $g ) { $f + $g } @array;

map :my( $f, $g ) $f + $g, @array;

map :my( $f ) { $f + 1 } @array;

map :my( $f ) $f + 1, @array;
```

leading colon is afaik a syntax error, and it would fit in with treating the block of the map as an anonymous subroutine that is gaining a signature, only in this case via something that looks like an attribute. parens mandatory to avoid any parsing ambiguity, no special asciis needed. visual separation fully maintained in all cases

leonerd:
@hvds 
> I don't really imagine I'd ever want to count the number of bundled elements, but if I needed it I'd multiply by n in the well-formed case and use something like `my $count = () = grep ...` for the trailing-undefs case.

Huh, I guess that is an equally-valid point. There's also the invariant that

```
my $count = grep { CODE... } ITEMS;
```

yields the number of times the block of code returned a true value. Perhaps that would be a more useful thing to keep in this case.

As I said I'm not hugely fussed either way, we should be careful to decide which we think is best and proceed accordingly.

wchristian:
the main question is the balance of usefulness vs consistency and surprise

```perl
# block count
my $count = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
say $count; ### 1
my @list = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
say 0 + @list; ### 2

# element count
my $count = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
say $count; ### 2
my @list = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
say 0 + @list; ### 2
```

the former is more useful but also exhibits some level of "magic", the latter would be easier to explain and reason about

i think it could be ok to go for useful as long as the matter of "it's another bit of weirdness to explain to newbies" is considered explicitly and consider to be a low amount of harm

rfcs/rfc-HAARG-0001.md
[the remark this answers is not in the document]
haarg: I would say that the scalar value should be the size of the list it would return.

The list being returned should be aliases of the incoming values. That is how grep works now. So while missing values should be undef during the iteration, they should be missing in the return value.

I'll write up some additions to cover this.

rfcs/rfc-HAARG-0001.md
haarg: My initial version of this idea did disable the EXPR form of map entirely, because I was concerned about this ambiguity. But @iabyn disagreed and said it wasn't needed. While the list of iteration variables could be unbounded, the syntax of them is simple. It doesn't require parsing perl's general and infinitely complex syntax to disambiguate, unlike the current BLOCK vs EXPR forms.

Having this feature disable the EXPR form of map would be decent compromise. But until someone proves that necessary, I don't think we need to change this in the RFC. It also means we don't need to come up with a name for the feature.

haarg:
In terms of usefulness, the answer of block count vs element count has an easy answer: the scalar return should be the element count, and if users want a block count they can use map rather than grep. And since this is behavior is less magic and easier to document, I think it's the clear best option.

rfcs/rfc-HAARG-0001.md
wchristian: > Having this feature disable the EXPR form of map would be decent compromise.

this would, on a long timescale, disable the expression form entirely for perl, due to social knock-on effects, and cause massive damage. it is not acceptable in any way

wchristian:
> if users want a block count they can use map rather than grep

how do you mean this? wouldn't one just divide the element count by the argument count?

rfcs/rfc-HAARG-0001.md
wchristian: also a compromise would be `:my(` syntax, as it makes the parsing wholly unambiguous and erases all aesthetic concerns about visual separation

(sorry for repost, bad github ui)

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,180 @@
...
+);
+
+my %pair_grep = grep my ($f, $g) { $f =~ /sec/ } %my_hash;
+# ( second_key => 14 )
+```
+
+## Prototype Implementation
+
+  - List::Util includes the functions `pairmap` and `pairgrep` which can do
+    n-at-a-time iteration. These use the "magic" `$a` and `$b` variables.
```
tobyink, line 127: Only assuming n=2.

leonerd: Yes; the pair* functions are "n-at-a-time" for specifically n=2; the ones suggested by this RFC are even more general than that.

leonerd:
@wchristian 
> i had other thoughts here, but that got derailed to hell by this suggestion:
> 
> > it might require a named feature to enable this syntax while simultaneously disabling the use of map EXPR, LIST
> 
> the downstream consequences of that would be very destructive. proposing that as a first line consideration instead of trying or even asking for compromises feels cartoonishly evil and selfish to me and is frankly making me want to give up on all of this if this is considered to be an acceptable way to treat users of perl. you can do better

These seem very strong words as compared my suggested change; I'm trying to understand where this is coming from.

Firstly just to clarify: You do appreciate that I mean that it would only disable `map EXPR, LIST` while that lexical feature was enabled in scope, yes? Not "totally remove it from perl" - that would indeed be crazy.

So, given that it is lexically opt-in, and you'd be asking for the new map/grep forms in general, I think it feels like a reasonable compromise to remove the EXPR form. But you seem to disagree.

Secondly then, I want to clarify what you feel it is we'd lose by not allowing the EXPR form. Yes it's one character more to type (well, three if you count the spaces), between

```perl
my @result = map { EXPR } LIST;
my @result = map EXPR, LIST;
```

but aside from that there really is no difference in behaviour. They both perform the exact same work in the exact same way. Indeed they even compile to the same ops in practice:

```
$ perl -MO=Concise -ce 'map EXPR, LIST'
8  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 1 -e:1) v:{ ->3
6     <|> mapwhile(other->7)[t1] vK ->8
5        <@> mapstart K ->6
3           <0> pushmark s ->4
-           <1> null lK/1 ->4
7              <$> const[PV "EXPR"] s/BARE ->6
4           <$> const[PV "LIST"] sM/BARE ->5
-e syntax OK

$ perl -MO=Concise -ce 'map { EXPR } LIST'
8  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 1 -e:1) v:{ ->3
6     <|> mapwhile(other->7)[t1] vK ->8
5        <@> mapstart K ->6
3           <0> pushmark s ->4
-           <1> null lK/1 ->4
-              <1> null lK/1 ->6
-                 <@> scope lK ->-
-                    <;> ex-nextstate(main 2 -e:1) v ->7
7                    <$> const[PV "EXPR"] s/BARE ->6
4           <$> const[PV "LIST"] sM/BARE ->5
-e syntax OK
```

Yes, the latter one did originally contain two extra OP_NULLs, an OP_SCOPE and an OP_NEXTSTATE, but observe that they have been nulled out. They don't appear in op execution order (`B::Concise` has not assigned a number to them in the lefthand column). During runtime the exact same ops will be executed in either case. They have identical performance.

Given, therefore, that these two different syntax forms behave identically, I don't understand why you feel so strongly that we must continue to offer users the choice to write their code in both forms, especially when we observe the complexity of distinguishing one form from the other during compiletime. As an author of a static analysis tool for Perl code, I would imagine you especially would understand the parser difficulties that result in doing so.

Before I try to convince you further, I'd like to hear your thoughts on this above. And in particular, I'd like to understand from you why you feel so strongly that `map EXPR, LIST` is a perl feature that we *must* continue to encourage users to use in newly-written code.

haarg:
> the main question is the balance of usefulness vs consistency and surprise
> 
> ```perl
> # block count
> my $count = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
> say $count; ### 1
> my @list = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
> say 0 + @list; ### 2
> 
> # element count
> my $count = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
> say $count; ### 2
> my @list = grep my ($k, $v) { $k eq "one" } ( one => 1, two => 2, three => 3 );
> say 0 + @list; ### 2
> ```
> 
> the former is more useful but also exhibits some level of "magic", the latter would be easier to explain and reason about
> 
> i think it could be ok to go for useful as long as the matter of "it's another bit of weirdness to explain to newbies" is considered explicitly and consider to be a low amount of harm

I don't think "usefulness vs consistency" is a useful way to think about this. The usefulness of returning the number of matches vs the number of elements doesn't have an obvious answer. Different use cases could call for different results. And no matter which we chose, the alternate option could be done by using a map. You could switch to map and return either one item to count matches or two matches to return element count. Doing math on the result could also work.

So I think we should ignore "usefulness" as unanswerable for this decision, and instead focus on consistency and ease of explanation. To me, that would point to scalar context returning the number of elements that would have been generated.

haarg:
Regarding some type of "lightweight lambda" for the map/grep iteration function, this is roughly what I was pointing to with my initial comments about being more sub-like. I don't think it makes sense to create something like this only for map/grep, but it could make sense as a more general feature. So I wouldn't support a syntax like any of:
```
map my $x onto $x ** 2, @numbers
map my $x ==> $x ** 2, @numbers
map my $x <- $x ** 2, @numbers
```
unless we were designing it as something usable across the entire language.

wchristian:
@leonerd thanks, i don't have the opportunity for a detailed response today, but i appreciate the way you engaged that and i think i can give useful answers for everything you asked

for now, please consider this as a demonstration of an issue which causes me to intentionally default to expression form in my code on some constructs (and choose the opposite for others if i can anticipate needing to inspect each step), which is also something i don't think can be fixed in the debugger

i'll be writing more tomorrow

```perl
Mithaldu@XENBOX /cygdrive/c/Users/Mithaldu/Videos
$ perl -d loop.pl

Loading DB routines from perl5db.pl version 1.57
Editor support available.

Enter h or 'h h' for help, or 'man perldebug' for more help.

main::(loop.pl:4):      my @things = generate_arbitrary_sized_list(); # might make 2, might make a million
  DB<1> n
main::(loop.pl:6):      my @new = map $_ + 1, @things;
  DB<1> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<1> x @new
0  2
1  3
2  4
3  5
4  6
5  7
6  8
7  9
  DB<2> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<2> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<2> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<2> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<2> x @slow_new
  empty array
  DB<3> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<3> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<3> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<3> n
main::(loop.pl:8):      my @slow_new = map { $_ + 1 } @things;
  DB<3> n
main::(loop.pl:10):     exit;
  DB<3> x @slow_new
0  2
1  3
2  4
3  5
4  6
5  7
6  8
7  9
  DB<4> n
Debugged program terminated.  Use q to quit or R to restart,
use o inhibit_exit to avoid stopping after program termination,
h q, h R or h o to get additional info.
  DB<4> q
```

@haarg
> I think we should ignore "usefulness" as unanswerable for this decision

my gut says for grep returning the number of matches will be most useful, but i'm happy with consistency in any case. no strong opinions there on my part. just wanted to spell out the possibilities clearly

leonerd:
@wchristian Any update yet?

While you're writing that, I think I can respond initially with more of a question. You point out that `map EXPR, LIST` vs `map {BLOCK} LIST` behave differently under `perl -d`. In particular that the debugger considers for purposes of the `n` or `s` commands that each execution of the block statement is its own statement, but the entire expression form is one statement.

When giving a further response, can you clarify with this, whether it's the case that you want to be able to write both kinds of loops (and in that case I'm interested why only `map` and `grep` are special here - why would you not care about e.g. being able to consider an entire `do {}` block to be one statement sometimes). Or is it the case that sometimes you'd want to step over these, and sometimes step into and see each iteration? If it's the latter then it sounds like the whole thing could be fixed in a far more useful and general way, by the debugger having better kinds of "next"-style command, being able to run the code "until some thing" for a variety of things. That way, on a per-command basis you could decide if you wanted to treat the bodies of `map` and `grep` as entire block to jump into or a single atomic thing to step over; but you could also do that for `if` blocks, `while` loops, `do` blocks, all other kinds of things.

Perhaps if the debugger had that ability, then you wouldn't need to distinguish expression- vs block-forms of `map`. But this is just me guessing - please feel free to elaborate further.

wchristian:
Work has been extreme this week, sorry.

That said: I'd appreciate a change to the debugger in that regard. It would be an improvement of the state of things in general because i don't always run code where i wrote the loop myself, and as you mentioned there are loops where it would be useful in other ways too.

However i would for other reasons still want to write both forms. And that's not only restricted to map and grep. With for i prefer the postfix form, but also still use both. On do i don't have an opinion because i personally don't use it much except for special cases where i want to hide created variables from the upper scope. Additionally, while i probably would like to see a term from of do, i wouldn't propose that in this specific RFC because it doesn't exist yet and would be a RFC entirely on its own, since it's at most slightly tangential to this one.

leonerd:
In case it's of interest to anyone, I've had a go at supporting this kind of syntax in my `List::Keywords` module. So now it supports one-at-a-time lexicals in `first` and `any/all/etc...`. I'll probably have a go at making an `ngrep` and `nmap` pair next.

leonerd:
Some extra thoughts that came up while discussing this:

* Perl's existing `foreach` loops will all alias the lexical to each element of the list, in the same way they do for global `$_`, making modifications possible and visible. For consistency `map` and `grep` should continue to with lexicals

* The additional `undef`s that n-at-a-time `foreach` sees are always immutable ones. This might matter for things like autovivification, because it means if you try to access elements of a referenced array or hash, you'll get errors:
```
foreach my($k, $v) (1) { $v->{key} = 5 }
for my (...) is experimental at - line 1.
Modification of a read-only value attempted at - line 1.
```

* The List::Util::pair* functions print a "Odd number of elements" warning when presented with a non whole-bundle sized input list. Should n-at-a-time map/grep do similar? foreach doesn't, it's silent in that case. Such a warning was never mentioned in the original n-at-a-time foreach RFC (0001), but perhaps this was simply overlooked, and it should?

leonerd:
Furthermore, I've now added an `ngrep` and `nmap` to List::Keywords.
https://metacpan.org/pod/List::Keywords#N-AT-A-TIME-FUNCTIONS

exodist:
I have not read the entire thread here, so not sure if this is relevent, but saw nmap and ngrep and decided to pop this out here in case it is relavent: https://metacpan.org/pod/Block::NamedVar

This used Devel::Declare and added named vars for grep and map (ngrep and nvar and nfor) that could use multiple arguments. Probably years out of date and may not work, but may be useful.

haarg:
It appears Block::NamedVar only supports using multiple variables on nfor, not nmap or ngrep. But I'll add it to the RFC as a reference.

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,204 @@
...
+
+### `map my (VAR, VAR) BLOCK LIST`
+
+This will evaluate `BLOCK` for each set of two elements in `LIST`, aliasing
+the `VAR` to the first of each set, and the second `VAR` to the second. More
+than two variables can be used to iterate over sets of three or more items.
+
+On the last iteration, there may not be enough elements remaining to fill
+every `VAR` slot. In this case, a warning will be issued, and the extra `VAR`
+slots will be filled with `undef`.
```
leonerd, line 90: The more I look around perl, the more I'm not sure if we'd want this warning or not. Other similar cases do not currently warn:

```perl
my ($x, $y, $z) = (1);

use feature 'foreach_list';
foreach my ($x, $y, $z) (1) { ... }
```

In fact, the only place where a warning turns up is in the special case of assigning a list to a hash:

```perl
eval: my %h = (1, 2, 3);
Odd number of elements in hash assignment at (eval 6) line 1.
```

but this is specific to hashes, not filling a list scalars.

rfcs/rfc-HAARG-0001.md
```diff
@@ -0,0 +1,204 @@
...
+### `grep my (VAR, VAR) BLOCK LIST`
+
+This will evaluate `BLOCK` for each set of two elements in `LIST`, aliasing
+the `VAR` to the first of each set, and the second `VAR` to the second. More
+than two variables can be used to iterate over sets of three or more items.
+
+On the last iteration, there may not be enough elements remaining to fill
+every `VAR` slot. In this case, a warning will be issued, and the extra `VAR`
+slots will be filled with `undef`. The extra `undef` values used will also be
+included in the returned list.
```
leonerd, line 103: Same comment as above re: warnings.

wchristian:
Sorry for taking a while.

You may remember that we had a conversation on IRC as well, where @shadowcat-mst provided substantial critique along lines i very much agreed with. That said, to the questions:

> @wchristian
> 
> > i had other thoughts here, but that got derailed to hell by this suggestion:
> > > it might require a named feature to enable this syntax while simultaneously disabling the use of map EXPR, LIST
> > 
> > the downstream consequences of that would be very destructive. proposing that as a first line consideration instead of trying or even asking for compromises feels cartoonishly evil and selfish to me and is frankly making me want to give up on all of this if this is considered to be an acceptable way to treat users of perl. you can do better
> 
> These seem very strong words as compared my suggested change; I'm trying to understand where this is coming from.
> 
> Firstly just to clarify: You do appreciate that I mean that it would only disable `map EXPR, LIST` while that lexical feature was enabled in scope, yes? Not "totally remove it from perl" - that would indeed be crazy.
>
> So, given that it is lexically opt-in, and you'd be asking for the new map/grep forms in general, I think it feels like a reasonable compromise to remove the EXPR form. But you seem to disagree.

Yes. I am talking about emergent consequences of taking that kind of path. Being able to lexically reenable it is useless. People will want to use this. It is extremely useful. I expect it will end up in a use x.x; pragma. At that point someone who works in a team and wants to use modern Perl but also expression form will have to either convince their team to build a custom pragma which disables the feature, which, being that it is super useful, would be a hard sell. Alternatively they can resort to this kind of thing (or similar) every time they write a map function, which i hope you would not need explained that it is just shit.

    ...;
    my @result;
    {
        no feature 'natatime';
        map whatever, @input;
    }

The practical end result of this, given the usefulness of the feature, is that in most teams the choice will fall in favor of the feature, and thus against *reasonable* accessibility of expression form map. In other words, nobody could use it other than by making their code completely unreasonable. And they also couldn't use the natatime feature together with it. And, as i will explain below and we discussed in IRC, this all affects the debuggability of code with map and grep.

> Secondly then, I want to clarify what you feel it is we'd lose by not allowing the EXPR form. Yes it's one character more to type (well, three if you count the spaces), between
> 
> ```perl
> my @result = map { EXPR } LIST;
> my @result = map EXPR, LIST;
> ```
> 
> but aside from that there really is no difference in behaviour. ... Given, therefore, that these two different syntax forms behave identically,

As we discussed on IRC, they do not behave identically under the debugger, which is a perl core feature. There are also accessibility concerns with permitting people to reduce their use of {} inputs on international keyboards in conjunction with joint health. And, since "i think it's ugly" counts as a whole and valid argument from you here: To me brace form is ugly and uncomfortable to read. To me expression form is faster to read.

> I don't understand why you feel so strongly that we must continue to offer users the choice to write their code in both forms,

In addition to the above concerns there are basic language design issues:

    21:38 <haarg> so map ends up with three distinct forms
    21:43 <mst> haarg: it already has two, so if you create a second dimension and
                add a case you're going to end up with four unless you fuck up the
                design, yes.
    22:24 <mst> I can see the theoretical underpinnings of that position but from a
                practical point of view and from an 'aesthetics of the language'
                point of view it feels to me (and, yes, now -I'm- giving an
                emotional response to a great extent) it's a position that's more
                python than perl
    22:26 <mst> also I feel like extending map but only half extending it is going
                to be a POLS violation as well as a DWIM violation
    22:27 <mst> I have no issue with "I'm not interested in implementing that part"
                but I do think whatever you do implement should leave room for
                somebody who does want to to do so - which is a much reduced level
                of thinking about it but I don't think ignoring it entirely will
                produce a good outcome

> especially when we observe the complexity of distinguishing one form from the other during compiletime. As an author of a static analysis tool for Perl code, I would imagine you especially would understand the parser difficulties that result in doing so.

As we discussed on IRC, that is solvable.

> Before I try to convince you further, I'd like to hear your thoughts on this above. And in particular, I'd like to understand from you why you feel so strongly that `map EXPR, LIST` is a perl feature that we _must_ continue to encourage users to use in newly-written code.

Incorrect question. [https://grep.metacpan.org/search?q=%5E%5B%5E%5C%23%5Cn%5D%2B%5B%5E%24%40%5D%5Cbmap%5Cs%2B%5B%5E%7B%5Cs%5D&qd=&qft=*.pm](As grep.metacpan.org/search) demonstrates, there is a vast demographic of people who willingly use the expression form. If you're implementing a feature for one half of map, it is upon you to provide an argument as to why you want to make the people who use the other half (whatever their reasons may be), choose between being able to use natatime or expression form.

A penultimate observation: Even haarg uses mostly expression form. As such i am frankly baffled as to why we're having this conversation in the first place.

    21:06 <haarg> most of my maps are map +(

And as a last word: While this RFC has one author and i can't make anyone do anything; the RFC process is for implementing language design changes. Language design changes cannot and must not be subject to the whims of purely personal preference. They must serve the users of Perl as a whole.

haarg:
This got a lot of discussion, and was given PPC#23, so I am planning to merge it in roughly its current form, keeping the exploratory status.

I'm not otherwise intending to pursue this at this time.

PR #34, opened 2023-02-26
Update README.md

jidanni:
Please mention something about how to tell if the email arrived. I am only making a guess as to what to write.


book:
Actually the spreadsheet (while being public) is mostly used as a work document by the PSC.
Progress happens on the GitHub repository and the p5p mailing list.

PR #35, opened 2023-03-27
Rename RFC to PPC

book:
As discussed by the Perl Steering Council, the term "RFC" is ambiguous, does not reflect the actual process, and already shows up in the Perl documentation with a different meaning.

We chose the term "PPC" (Proposed Perl Change) to replace it. It is closer to the actual process (people _propose_ changes, which might be accepted and eventually implement, or not), and avoids any confusion with the venerable IETF documents.

No acronym is unique, and this one collides with "PowerPC" and "Pay-Per-Click". However, the defunct CPU architecture and the web marketing term are unlikely to show up in the Perl documentation and create any ambiguity.


rwp0 (approved):

rjbs (approved):
Let's do it…

PR #38, opened 2023-07-28
ppc0017: Change status to Implemented

rwp0:
Implemented within the "builtin" namespace as below:

https://perldoc.perl.org/blead/builtin#created_as_string

https://perldoc.perl.org/blead/builtin#created_as_number


leonerd (approved):
LGTM

PR #39, opened 2023-07-28
ppc0018: Change status to Implemented

rwp0:
Implemented as the "module_true" feature:

https://perldoc.perl.org/blead/feature#The-'module_true'-feature


leonerd (approved):
LGTM

PR #43, opened 2024-01-02
ppc0013: Use `List::Util` for `reduce`

rwp0:
It also fixes GitHub's Markdown's
highlighting the Perl code correctly

*it seems if there's # (comments) above
the code it fails to recognize ```perl*

Bump 5.34 to 5.38 since there's still
significant progress in this field until now

### DIFF
--- a/ppcs/ppc0013-overload-join-and-substr.md
+++ b/ppcs/ppc0013-overload-join-and-substr.md
@@ -9,7 +9,7 @@
 
 ## Abstract
 
-As of Perl version 5.34, `overload` is incomplete and surprising for
+As of Perl version 5.38, `overload` is incomplete and surprising for
 string operations:
 
 * the `join` builtin should be using the `concat` (`.`) overload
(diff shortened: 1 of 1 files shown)


rwp0:
Surprised to meet `reduce` which is not in `CORE::`.

ppcs/ppc0013-overload-join-and-substr.md
```diff
@@ -98,6 +98,8 @@ We do not yet foresee security issues. Guidance is welcome.
 For `join`:
 
 ```perl
+use List::Util;
```
Grinnz, line 101: reduce must be imported explicitly from List::Util, there are no default exports.

rwp0: Thanks will rework

rwp0: Done @Grinnz thanks for the feedback!

ppcs/ppc0013-overload-join-and-substr.md
```diff
@@ -98,11 +98,14 @@ We do not yet foresee security issues. Guidance is welcome.
...
-# we expect this code:
+use List::Util 'reduce';
+
+# When @list contains elements with concat overloading, we expect this code:
+
 my $ret = join $sep, @list;
 
-# to behave like this code:
+# To behave like this code:
+
```
leonerd, line 108: As it's only relevant to this line, surely put the the `use List::Util 'reduce';` down here?

rwp0: I see, yes, force-pushed.

rwp0:
Can we merge this now?

rwp0:
Any updates on this? Maybe I need to change more?

PR #50, opened 2024-07-17
Adds a PPC for `any` and `all` as built-in operator syntax

leonerd:
A reminder on the PPC process: Accepting this PR means we like the document, but it **does not** automatically imply that we accept the idea embodied by it. Accepting the document first is useful as it gives it a number and allows us to discuss and iterate on one canonical version of it, ahead of accepting the underlying idea. It is perfectly fine to accept the document even if questions and issues still surround it.

### DIFF
--- a/ppcs/ppcTODO-more-list-utils.md
+++ b/ppcs/ppcTODO-more-list-utils.md
@@ -29,8 +29,6 @@ all { BLOCK } LIST
 
 These operators are similar to `grep` in scalar context, though yield a simple boolean truth value relating to how many input values made the filter block yield true. `any` yields true when its block yields true for at least one of its input values, or false if they all yield false. `all` yields true only if every input value makes the block yield true, or false if at least one yields false.
 
-These operators only yield a single scalar; in list context therefore they will just provide a single-element list containing that boolean scalar.
-
 A consequence of these rules is what happens when given an empty list. A call to `any` with an empty list does not have any input values which made the block return true, so its result is false. Conversely, a call to `all` with an empty list does not have any input values which made the block return false, so its result is true.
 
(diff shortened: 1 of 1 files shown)


ppcs/ppcTODO-more-list-utils.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+any { BLOCK } LIST
+
+all { BLOCK } LIST
+```
+
+These operators are similar to `grep` in scalar context, though yield a simple boolean truth value relating to how many input values made the filter block yield true. `any` yields true when its block yields true for at least one of its input values, or false if they all yield false. `all` yields true only if every input value makes the block yield true, or false if at least one yields false.
+
+A consequence of these rules is what happens when given an empty list. A call to `any` with an empty list does not have any input values which made the block return true, so its result is false. Conversely, a call to `all` with an empty list does not have any input values which made the block return false, so its result is true.
```
rabbiveesh, line 32: I was at first surprised by this behavior, that all would be true on an empty list; though i did see that that's how `every` works in javascript and `all` in python, and i suppose it makes sense. I think we could have a word on why this design choice makes sense from a user's perspective rather than just the implementation reason.

book:
Why guard them by a feature, instead of adding them as `builtin` (and adding them to a later builtin version bundle)?

Leont:
> Why guard them by a feature, instead of adding them as builtin (and adding them to a later builtin version bundle)?

Because these aren't quite normal functions, the same way grep isn't.

Grinnz:
When builtin was being designed, at one point a second "keywords" namespace was proposed (possibly by me) for keywords that may have unique parsing behavior. But feature handles these just fine as long as we don't mind the extra feature proliferation.

book:
Thanks @Leont and @Grinnz for the explanation.

PR #51, opened 2024-07-19
Initial attempt at writing a PPC for custom prefix and postfix operators

leonerd:
A reminder on the PPC process: Accepting this PR means we like the document, but it **does not** automatically imply that we accept the idea embodied by it. Accepting the document first is useful as it gives it a number and allows us to discuss and iterate on one canonical version of it, ahead of accepting the underlying idea. It is perfectly fine to accept the document even if questions and issues still surround it.

In particular, given the lack of enthusiasm in this idea thus far, it may be worth accepting the PR purely to have a place to discuss whether it is even worth doing at all, accepting that we may ultimately end up rejecting it. At least this will serve as a discoverable place to document why.

### DIFF
--- a/ppcs/ppcTODO-custom-prefix-postfix-operators.md
+++ b/ppcs/ppcTODO-custom-prefix-postfix-operators.md
@@ -1,0 +1,127 @@
+# Custom Pre- and Post-fix Operators
+
+## Preamble
+
+    Author:  Paul Evans <PEVANS>
+    Sponsor:
+    ID:      TODO
+    Status:  Exploratory
+
+## Abstract
+
+Expand the current `PL_infix_plugin` and mechanism and `XS::Parse::Infix` to support prefix and postfix operators as well.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


book:
If I remember correctly, you said on p5p something to the effect of "writing this PPC conviced me it wasn't worth it".

Should we close the PR as it is?

leonerd:
I don't think we should close it entirely. It's still a useful suggestion of how to add such a feature *if* there was a compelling reason to do so. Perhaps instead I should adjust the wording to say "at the moment we don't have a strong enough case to actually do this, so it remains unimplemented".

ajmetz:
I was gonna say - isn't custom pre and post stuff available via chaining custom methods, which basically is akin to the "existing function-call notation" mentioned in the PPC. That said, your example of `√($x² + $y²)` is very cool and neat, even if you immediately show the same can currently be done with `sqrt($x**2 + $y**2)` and there isn't enough justification yet for implementing this PPC. I must try out the cool infix stuff at some point though, =).

PR #54, opened 2024-08-15
Signature named params

leonerd:
Another PPC doc to add to the growing pile ;)

As is traditional for me I didn't fill in the "Rationale" section as I still have no idea how that differs from a "Motivation". The two words seem rather synonymous to me.

A reminder on the PPC process: Accepting this PR means we like the document, but it **does not** automatically imply that we accept the idea embodied by it. Accepting the document first is useful as it  gives it a number and allows us to discuss and iterate on one canonical version of it, ahead of accepting the underlying idea. It is perfectly fine to accept the document even if questions and issues still surround it.

### DIFF
--- a/ppcs/ppcTODO-signature-named-parameters.md
+++ b/ppcs/ppcTODO-signature-named-parameters.md
@@ -42,6 +42,20 @@ Each element provides a new lexical variable that is visible during the body of 
 
 The value of a named parameter is taken from the argument values passed by the caller, in a manner familiar to existing uses of hash assignment. The caller should pass an even-sized name-value pair list. The values corresponding to names of parameters will be assigned into the variables. The order in which the values are passed by the caller is not significant.
 
+Since it is a relatively common pattern in callsites in existing code to rely on the semantics of assignment of name-value pair lists into hashes, the beahviour on encountering duplicate key names needs to be preserved. This is that duplicated key names do not raise an error or a warning, and simply accept the last value associated with that name. This allows callers to collect values from multiple sources with different orders of priority to override them; for example using a hash of values combined with individual elements:
+
+```perl
+sub func ( :$abc, :$xyz, ... ) { ... }
+
+func(
+    abc => 123,
+    %args,
+    xyz => 789,
+);
+```
+
+In this example, the given `abc` value will take effect unless overridden by a later value in `%args`, but the given `xyz` value will replace an earlier one given in `%args`. Neither situation will result in a warning or error.
+
 Furthemore, all of the new behaviour is performed within the body of the invoked subroutine entirely by inspecting the values of the arguments that were passed. The subroutine is not made aware of how those values came to be passed in - whether from literal name-value syntax, a hash or array variable expansion, or any other expression yielding such a list of argument name and value pairs.
 
 ```perl
@@ -65,6 +79,8 @@ make_colour( red => 1.0, blue => 0.5 );
 #   $green = 0
 #   $blue  = 0.5
 ```
+
+Since defaulting expressions are full expressions and not necessarily simple constant values, the time at which they are evaluated is significant. Much like with positional parameters, each is evaluated in order that it is written in source, left to right, and each can make use of values assigned by earlier expressions. This means they are evaluated in the order that is written in the function's declaration, which may not match the order that values were passed from the caller in the arguments list.
 
 A subroutine is permitted to use a combination of *mandatory* positional and named parameters in its definition, provided that all named parameters appear after all the positional ones. Any named parameters may be optional; there are no ordering constraints here.
 
--- a/ppcs/ppcTODO-signature-named-parameters.md
+++ b/ppcs/ppcTODO-signature-named-parameters.md
@@ -86,7 +86,49 @@ There are no anticipated security concerns with expanding the way that subroutin
 
 As the intention of this syntax addition is to make existing code practice more consise and simple to write, it would be illustrative to compare pairs of functions written in the newly proposed vs. the existing style.
 
-TODO
+One example comes from [`IPC::MicroSocket::Server`](https://metacpan.org/pod/IPC::MicroSocket::Server). This is currently using the `Sublike::Extended` module (see "Prototype Implementation" below) to provide the syntax. The actual module also uses `Object::Pad` to provide the `method` keyword; this example is paraphrased to avoid that. This module requires a `path` named argument, and optionally takes a `listen` argument, defaulting its value to 5 if the caller did not provide a defined value.
+
+```perl
+extended sub new_unix ( $class, :$path, :$listen //= 5 )
+{
+    ...
+}
+```
+
+This replaces the previous version of the code which handled arguments by the more traditional approach of assigning into a hash:
+
+```perl
+sub new_unix ( $class, %args )
+{
+    my $path   = $args{path};
+    my $listen = $args{listen} // 5;
+    ...
+}
+```
+
+Already the new code is shorter and more consise. Additionally it contains error checking that complains about missing mandatory keys, or unrecognised keys, which the previous version of the code did not include.
+
+Another example, this time from [`Text::Treesitter::QueryCursor`](https://metacpan.org/pod/Text::Treesitter::QueryCursor). This method takes an optional argument named `multi`, tested for truth. If absent it should default to false.
+
+```perl
+sub next_match_captures ( $self, :$multi = 0 )
+{
+    ...
+}
+```
+
+The previous version of this code did include complaints about unrecognised keys, and was rather longer because of it:
+
+```perl
+sub next_match_captures ( $self, %options )
+{
+    my $multi = delete $options{multi};
+    keys %options and
+        croak "Unrecognised options to ->next_captures: " . join( ", ", keys %options );
+
+    ...
+}
+```
 
 ## Prototype Implementation
 
--- a/ppcs/ppcTODO-signature-named-parameters.md
+++ b/ppcs/ppcTODO-signature-named-parameters.md
@@ -134,6 +134,16 @@ sub next_match_captures ( $self, %options )
 
 The [`XS::Parse::Sublike`](https://metacpan.org/pod/XS::Parse::Sublike) module contains parsing to allow third-party syntax modules to parse subroutine-like constructions, and includes a parser for named parameters already having this syntax. These are also made available to regular subroutines via the `extended` keyword provided by [`Sublike::Extended`](https://metacpan.org/pod/Sublike::Extended).
 
+Additionally, other existing CPAN modules already parse syntax in this, or a very similar format:
+
+* [`Function::Parameters`](https://metacpan.org/pod/Function::Parameters)
+
+* [`Kavorka`](https://metacpan.org/dist/Kavorka/view/lib/Kavorka/Manual/Signatures.pod)
+
+* [`Method::Signatures`](https://metacpan.org/pod/Method::Signatures)
+
+All of these use the leading-colon syntax in a signature declaration to provide named parameters in the same style as this proposal. It would appear we are in good company here.
+
 ## Future Scope
 
 * If the Metaprogramming API (PPC0022) gains introspection abilities to enquire about subroutine signature parameters, further consideration will need to be made in that API on how to represent the extra kinds of parameters added by this specification.
--- a/ppcs/ppcTODO-signature-named-parameters.md
+++ b/ppcs/ppcTODO-signature-named-parameters.md
@@ -154,7 +154,29 @@ All of these use the leading-colon syntax in a signature declaration to provide 
 
 ## Open Issues
 
-* How to specify a named parameter whose name is anything other than the name of the lexical variable into which its value is assigned? This is more of note when considering that traditional argument handling techniques involving assignment into hashes can handle more styles of name that would be invalid for lexical variables - for example, names including hyphens.
+### Parameter Names
+
+How to specify a named parameter whose name is anything other than the name of the lexical variable into which its value is assigned? This is more of note when considering that traditional argument handling techniques involving assignment into hashes can handle more styles of name that would be invalid for lexical variables - for example, names including hyphens.
+
+Perhaps a solution would be to permit attributes on parameter variables, then define a `:name` attribute for this purpose.
+
+```perl
+sub display ( $message, :$no_colour :name(no-colour) ) {
+    ...
+}
+```
+
+Further thoughts in that direction suggested that this could also support multiple names with aliases:
+
+```perl
+sub display ( $message, :$no_colour :name(no-colour no-color) ) {
+    ...
+}
+```
+
+The attribute syntax starting with a leading colon does visiually look quite similar to the named parameter syntax which also uses. This is a little unfortunate, but perhaps an inevitable consequence of the limited set of characters available from ASCII. As attributes are already well-established as leading with a colon, the only other option would be to pick a different character for named attributes; but this would be contrary to the established convention of the existing modules listed above.
+
+Perl does not currently support attributes being applied to parameters in signatures, named or otherwise, but this is the subject of a future PPC document I am currently drafting. I will add a reference here when it is published.
 
 ## Copyright
 
--- a/ppcs/ppcTODO-signature-named-parameters.md
+++ b/ppcs/ppcTODO-signature-named-parameters.md
@@ -1,0 +1,111 @@
+# Named Parameters in Signatures
+
+## Preamble
+
+    Author:  Paul Evans <PEVANS>
+    Sponsor:
+    ID:      TODO
+    Status:  Exploratory
+
+## Abstract
+
+Adds the ability for subroutine signatures to process named arguments in the form of name-value pairs passed by the caller, in a fashion familiar to existing uses of assignment into a hash.
+
+## Motivation
+
+Perl 5.20 added "subroutine signatures", native syntax for more succinctly expressing the common patterns of processing arguments inbound into a subroutine by unpacking the `@_` array. Rather than writing common styles of code directly, a more compact notation, in a style immediately familiar to users of many other languages, allows Perl to create the required behaviour from that specification.
+
+```perl
+sub f ($x, $y, $z = 123) { ... }
+
+# instead of
+sub f {
+    die "Too many arguments" if @_ > 3;
+    die "Too few arguments" if @_ < 2;
+    my ($x, $y, $z) = @_;
+    $z = 123 if @_ < 3;
+    ...
+}
+```
+
+One common form of argument processing involves passing an even-sized list of key/value pairs and assigning that into a hash within the subroutine, so that specifically named parameters can be extracted from it. There is currently no support from subroutine signature syntax to assist authors in providing such behaviours.
+
+## Rationale
+
+(explain why the (following) proposed solution will solve it)
+
+## Specification
+
+A new kind of element may be present in a subroutine signature, which consumes a named argument from the caller. These elements are written with a leading colon prefix (`:$name`), indicating that it is named rather than positional. The name of each parameter is implied by the name of the lexical into which it is assigned, minus the leading `$` sigil.
+
+Each element provides a new lexical variable that is visible during the body of the function, in the same manner as positional ones.
+
+The value of a named parameter is taken from the argument values passed by the caller, in a manner familiar to existing uses of hash assignment. The caller should pass an even-sized name-value pair list. The values corresponding to names of parameters will be assigned into the variables. The order in which the values are passed by the caller is not significant.
+
+Furthemore, all of the new behaviour is performed within the body of the invoked subroutine entirely by inspecting the values of the arguments that were passed. The subroutine is not made aware of how those values came to be passed in - whether from literal name-value syntax, a hash or array variable expansion, or any other expression yielding such a list of argument name and value pairs.
+
+```perl
+sub make_colour ( :$red, :$green, :$blue ) { ... }
+
+make_colour( red => 1.0, blue => 0.5, green => 0.2 );
+# The body of the function will be invoked with
+#   $red   = 1.0
+#   $green = 0.2
+#   $blue  = 0.5
+```
+
+As with positional parameters, a named parameter without a defaulting expression is mandatory, and an error will be raised as an exception if the caller fails to pass a corresponding value. A defaulting expression may be specified using any of the operators available to positional parameters - `=`, `//=` or `||=`.
+
+```perl
+sub make_colour ( :$red = 0, :$green = 0, :$blue = 0 ) { ... }
+
+make_colour( red => 1.0, blue => 0.5 );
+# The body of the function will be invoked with
+#   $red   = 1.0
+#   $green = 0
+#   $blue  = 0.5
+```
+
+A subroutine is permitted to use a combination of *mandatory* positional and named parameters in its definition, provided that all named parameters appear after all the positional ones. Any named parameters may be optional; there are no ordering constraints here.
+
+If a subroutine uses named parameters then it may optionally use a slurpy hash argument as its final element. In this case, the hash will receive all *other* name-value pairs passed by the caller, apart from those consumed by named parameters. If the subroutine does not use a slurpy argument, then it will be an error raised as an exception for there to be any remaining name-value pairs after processing.
+
+While all of the above has been specified in terms of subroutines, every point should also apply equally to the methods provided by the `class` feature, after first processing the implied invocant `$self` parameter.
+
+## Backwards Compatibility
+
+At the site of a subroutine's definition, this specification only uses new syntax in the form of the leading colon prefix on parameter names. Such syntax was formerly invalid in previous versions of perl. Thus there is no danger of previously-valid code being misinterpreted.
+
+All of the behaviour provided by this new syntax is compatible with and analogous to any existing code that could have been written prior, perhaps by direct assignment into a hash. There are no visible differences in the external interface to a subroutine using such syntax, and it remains callable in exactly the same way. Functions provided by modules could be upgraded to use the new syntax without any impact on existing code that invokes them.
+
+## Security Implications
+
+There are no anticipated security concerns with expanding the way that subroutines process parameters in this fashion.
+
+## Examples
+
+As the intention of this syntax addition is to make existing code practice more consise and simple to write, it would be illustrative to compare pairs of functions written in the newly proposed vs. the existing style.
+
+TODO
+
+## Prototype Implementation
+
+The [`XS::Parse::Sublike`](https://metacpan.org/pod/XS::Parse::Sublike) module contains parsing to allow third-party syntax modules to parse subroutine-like constructions, and includes a parser for named parameters already having this syntax. These are also made available to regular subroutines via the `extended` keyword provided by [`Sublike::Extended`](https://metacpan.org/pod/Sublike::Extended).
+
+## Future Scope
+
+* If the Metaprogramming API (PPC0022) gains introspection abilities to enquire about subroutine signature parameters, further consideration will need to be made in that API on how to represent the extra kinds of parameters added by this specification.
+
+## Rejected Ideas
+
+* This specification intentionally makes no changes to the call-site of any subroutines using named parameters.
+
+## Open Issues
+
+* How to specify a named parameter whose name is anything other than the name of the lexical variable into which its value is assigned? This is more of note when considering that traditional argument handling techniques involving assignment into hashes can handle more styles of name that would be invalid for lexical variables - for example, names including hyphens.
+
+## Copyright
+
+Copyright (C) 2024, Paul Evans.
+
+This document and code and documentation within it may be used, redistributed and/or modified under the same terms as Perl itself.
--- a/ppcs/ppc0024-signature-named-parameters.md
+++ b/ppcs/ppc0024-signature-named-parameters.md
@@ -4,7 +4,7 @@
 
     Author:  Paul Evans <PEVANS>
     Sponsor:
-    ID:      TODO
+    ID:      0024
     Status:  Exploratory
 
 ## Abstract


ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,153 @@
...
+Each element provides a new lexical variable that is visible during the body of the function, in the same manner as positional ones.
+
+The value of a named parameter is taken from the argument values passed by the caller, in a manner familiar to existing uses of hash assignment. The caller should pass an even-sized name-value pair list. The values corresponding to names of parameters will be assigned into the variables. The order in which the values are passed by the caller is not significant.
+
+Furthemore, all of the new behaviour is performed within the body of the invoked subroutine entirely by inspecting the values of the arguments that were passed. The subroutine is not made aware of how those values came to be passed in - whether from literal name-value syntax, a hash or array variable expansion, or any other expression yielding such a list of argument name and value pairs.
+
+```perl
+sub make_colour ( :$red, :$green, :$blue ) { ... }
+
+make_colour( red => 1.0, blue => 0.5, green => 0.2 );
```
happy-barney, line 50: let this syntax to use sigils, it may later allow you to use multiple lists
such syntax may also allow developer to use named syntax for positional parameters

leonerd: That's entirely against the whole point. The whole point as the document states, is that the callsites for these functions do not need to be changed or know about it. Much like with existing signature syntax, these extensions are simply better ways that the *body* of the function could be implemented, without needing the caller of the function to know about it.

happy-barney: hm, you are most likely right, with existing code and current state of syntax and parser it may be difficult to determine state I'd like to have without prefix token. FYI, I was thinking about function like one in `Net::Amazon::S3`, for example `add_key`, which you can call as:
````
add_key ($bucket, $key);
add_key (bucket => $bucket, key => $key);
add_key ($key, bucket => $bucket);
add_key ($key, name => $bucket); # named parameter bucket has an alias
```

JRaspass:
You say named parameters must appear after positional parameters but are we still able to put a slurpy unnamed hash at the end to make this a drop-in safe replacement for existing code in the same way that adding a slurpy unnamed array makes existing subroutine support a drop-in replacement:
```perl
sub foo { my ( $bar, $baz ) = @_ ) { .. }
# ↓
sub foo( $bar, $baz, @ ) { ... }
```
```perl
sub foo { my %args = @_; ... }
# ↓
sub foo( :$bar, :$baz, % ) { ... }
```

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,153 @@
...
+
+* If the Metaprogramming API (PPC0022) gains introspection abilities to enquire about subroutine signature parameters, further consideration will need to be made in that API on how to represent the extra kinds of parameters added by this specification.
+
+## Rejected Ideas
+
+* This specification intentionally makes no changes to the call-site of any subroutines using named parameters.
+
+## Open Issues
+
+* How to specify a named parameter whose name is anything other than the name of the lexical variable into which its value is assigned? This is more of note when considering that traditional argument handling techniques involving assignment into hashes can handle more styles of name that would be invalid for lexical variables - for example, names including hyphens.
```
mauke, line 147: Another issue is that subroutines can't currently check whether named parameters with default values had something passed in by the caller. That is, if you do
```perl
sub foo(:$arg => 42) {
    ...
}
```
then the two calls `foo()` and `foo(arg => 42)` are indistinguishable by `foo`.

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,153 @@
...
+
+While all of the above has been specified in terms of subroutines, every point should also apply equally to the methods provided by the `class` feature, after first processing the implied invocant `$self` parameter.
+
+## Backwards Compatibility
+
+At the site of a subroutine's definition, this specification only uses new syntax in the form of the leading colon prefix on parameter names. Such syntax was formerly invalid in previous versions of perl. Thus there is no danger of previously-valid code being misinterpreted.
+
+All of the behaviour provided by this new syntax is compatible with and analogous to any existing code that could have been written prior, perhaps by direct assignment into a hash. There are no visible differences in the external interface to a subroutine using such syntax, and it remains callable in exactly the same way. Functions provided by modules could be upgraded to use the new syntax without any impact on existing code that invokes them.
```
mauke, line 79: This assumes named parameters use hash initializer semantics:

 - Duplicate keys are allowed in calls.
 - The last instance of a particular key "wins" (but the preceding arguments are still evaluated).

leonerd:
> You say named parameters must appear after positional parameters but are we still able to put a slurpy unnamed hash at the end to make this a drop-in safe replacement for existing code in the same way that adding a slurpy unnamed array makes existing subroutine support a drop-in replacement:

It says exactly that:

> If a subroutine uses named parameters then it may optionally use a slurpy hash argument as its final element. In this case, the hash will receive all *other* name-value pairs passed by the caller, apart from those consumed by named parameters.

mauke (approved):

ppcs/ppcTODO-signature-named-parameters.md
leonerd: Imean... Yes, that's true, but if such a distinction is actually meaningful to the function then I'd suggest "don't do that". If you need to distinguish a 42 being given from the caller vs 42 from your own default, then you need to rethink something. Maybe just take an undef default, check it for being defined into some flag, and then `//= 42` it anyway..?

ppcs/ppcTODO-signature-named-parameters.md
leonerd: Mmm. This is fair; there could be other equally-valid implementations. The implementation could warn about duplicate keys, pointing out that earlier values were overwritten by later ones.

It could even be left as an extension idea to e.g. define a ":collect" attribute that makes an array of all the values given to a particular arg:

```perl
sub f ( :@x :collect ) { say "Value $_" for @x; }

f( x => 123, x => 456, x => 789 );
```

It seems fairly niche interest at that point.

Of course, nothing stops a function from simply taking a `'@rest` array and unpacking it themselves; perhaps in a multivalue foreach loop. The aim of this PPC is to try to find a balance of making the common cases easy, without getting in the way of the weird ones.

ppcs/ppcTODO-signature-named-parameters.md
xenu: I could think of some use-cases, for example code that inserts something to a database. `undef` is `NULL`, while no value means "insert default value".

Of course this could be handled manually, with a slurpy hash parameter.

Hmm... what if exists() could be used on named arguments?

xenu: PS. I agree it's a niche case and it's probably not worth worrying about.

ppcs/ppcTODO-signature-named-parameters.md
perigrin: The first thing I saw with `sub f(:@x :collect) {...}` is that I can't quickly tell that someone didn't make a typo meaning `sub f(:@x, :$collect) {...}`:

```perl
sub f(:@x, :$collect) { 
     say "x: $_" for @x;

     # 400+ lines of code

     $payment_processor->get_money(@x) if $collect;
}

f( x => 123, x => 456, x => 789, collect => true );
```

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+## Prototype Implementation
+
+The [`XS::Parse::Sublike`](https://metacpan.org/pod/XS::Parse::Sublike) module contains parsing to allow third-party syntax modules to parse subroutine-like constructions, and includes a parser for named parameters already having this syntax. These are also made available to regular subroutines via the `extended` keyword provided by [`Sublike::Extended`](https://metacpan.org/pod/Sublike::Extended).
+
+## Future Scope
+
+* If the Metaprogramming API (PPC0022) gains introspection abilities to enquire about subroutine signature parameters, further consideration will need to be made in that API on how to represent the extra kinds of parameters added by this specification.
+
```
perigrin, line 140: Is there potentially scope for using this outside of function params to de-structure paired lists in other contexts?

```perl
my @users = ( 
    { user => 'bob',    pass => 'abc123' },
    { user => 'alice',  pass => 'abc123' },
    { user => 'ted',    pass => 'abc123' },
    { user => 'caroll', pass => 'abc123' },
);

for my $user_entry (@users) { 
   my (:$user, :$pass) = $_->%*;
   say "$user password is $pass";
}
```
Or
```perl
use List::Util qw(mesh);

sub map_users_to_accounts(@users) { 
   my @accounts = get_accounts_for(@users);
   return mesh \@users, \@accounts;
}

my (:$bob, :$alice, :$ted, :$caroll) = map_users_to_accounts(@users);
```
And can we use it in the opposite direction?
```perl
my $path = get_path();
new_unix( $class, :$path )
```
and 
```perl
my ($bob, $alice, $ted, $caroll) = get_users();
my %accounts = (:$bob, :$alice, :$ted, :$caroll);
```

ppcs/ppcTODO-signature-named-parameters.md
happy-barney: @xenu you can still start to use `%_` to store only arguments set by caller so to distinguish set value and default value you can easily do:
```
if (exists $_{arg})
```

ppcs/ppcTODO-signature-named-parameters.md
happy-barney: Test case suggestion:
```
sub foo (:$x, @x);
foo (x => 1, x => q (slurpy));
```

ppcs/ppcTODO-signature-named-parameters.md
rabbiveesh: we had discussed this on the mailing list many moons ago, that allowing users to use named syntax for positional parameters is not good behavior by default b/c it suddenly exposes your internal API as external (b/c the names you happen to have chosen as params now matter at the call site).

ppcs/ppcTODO-signature-named-parameters.md
rabbiveesh: i second this idea, tho i think it's out of scope for this PPC persay; it's good to note that the design here is forward compatible with this type of syntax so maybe we should leave a note about that

leonerd: Certainly there is, as a separate idea. In fact I've been working on a `Syntax::Operator::Assign`, not yet on CPAN, to do exactly that kind of thing. Its LHS is supposed to accept anything that a signature would allow; in effect it means that function invocation is just an assignment from the caller's arguments into the signature list. Having named parameters in signatures is definitely a stepping-stone towards that, among other ideas.

ppcs/ppcTODO-signature-named-parameters.md
EvanCarroll: > we had discussed this on the mailing list many moons ago, that allowing users to use named syntax for positional parameters is not good behavior by default b/c it suddenly exposes your internal API as external (b/c the names you happen to have chosen as params now matter at the call site).

Is this an argument unique to the implementation of named params here, or are you arguing that named params in the abstract are a bad idea?

rabbiveesh: > > we had discussed this on the mailing list many moons ago, 
> 
> Is this an argument unique to the implementation of named params here, or are you arguing that named params in the abstract are a bad idea?

The argument I made was against the global call-site syntax that happy-barney recommended for named parameters. That has the issue of exposing your internal API, as mentioned. Named params in general are fine if you decide that's what your API should look like.

happy-barney: @rabbiveesh 

regarding: positional vs name call syntax. That definitely should be up to caller. Most likely caller is aware of context where function is called therefor can adjust call syntax accordingly. Function author should not dictate how to call it, only what constraint applies.

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,153 @@
...
+```perl
+sub make_colour ( :$red, :$green, :$blue ) { ... }
+
+make_colour( red => 1.0, blue => 0.5, green => 0.2 );
+# The body of the function will be invoked with
+#   $red   = 1.0
+#   $green = 0.2
+#   $blue  = 0.5
+```
```
oodler577, line 55: It's hard for me to follow, but does this not give you the unpack equivalent of `my %param = @_` when unambiguous?

tm604: If you're suggesting that this should give `$param = { red => 1.0, green => 0.2, blue => 0.5 }`, that'd make this entire thing PPC pointless: why not just `sub make_colour (%param)` in that case? You can already do that now, but it doesn't provide typo protection (`$param{bleu}`) and everything now has to go through that extra hash indirection - more typing involved, closures end up with the entire hash of parameters instead of the individual values, etc.

If the suggestion is that the _callsite_ should be a hashref, that adds extra syntax and I think makes the code less clear: why would named parameters need to be a hashref instead of a list? It's not going to end up in a hashref in the sub body.

oodler577: Then this whole thing is pointless. When anyone is passing in named parameters with `=>`, shoving them into a litany of scalars is forcing them in the baby Perl tier; I'm willing to bet 99% of the time the author wants a HASH ref.

Grinnz: The common presence of this idiom across CPAN and constant requests for the feature would suggest it is in fact not pointless.

happy-barney: > Then this whole PPC is pointless. When anyone is passing in named parameters with `=>`, shoving them into a litany of scalars is forcing them in the baby Perl tier; I'm willing to bet 99% of the time the author wants a HASH ref to keep things organized in the `sub`. It's a bait-n-switch. You're guaranteeing anyone who has any sense of idiomatic Perl will _not_ use this. If there are more advanced cases where this actually pays off, please show them. Because that RBG example is extremely lame.

I partially agree with you, especially when using slurpy hash, which may contain same keys as signature parameters.
Personally I think named parameter call syntax should not collide with any other syntax, which leads to two available solution:
- prefix named parameters with token
- prefix slurpy parameters with token (similar to `--` recognized by many command line tools)

Just returning to my earlier proposal:
```
make_colour ($r, $b, := $green => ...);
```

(which allows passing multiple lists as well has possible benefits beyond passing just variables)

tm604: > Then this whole PPC is pointless. When anyone is passing in named parameters with `=>`, shoving them into a litany of scalars is forcing them in the baby Perl tier; I'm willing to bet 99% of the time the author wants a HASH ref to keep things organized in the `sub`. It's a bait-n-switch. You're guaranteeing anyone who has any sense of idiomatic Perl will *not* use this. If there are more advanced cases where this actually pays off, please show them. Because that RBG example is extremely lame.

I gave you specific examples already, though?

- compiletime typo protection
- not having to keep an entire hash of unrelated things around in closures

If you don't use scalars anywhere and just keep everything in hashrefs, that's slightly odd, but your choice - extracting out to scalars is very much idiomatic Perl, though. If you've never seen `my ($x, $y) = @_` or `my $x = delete $args{x} // 'default';` then you're in a very different Perl world, and perhaps this PPC just isn't relevant to you.

rabbiveesh: > Then this whole PPC is pointless. When anyone is passing in named parameters with `=>`, shoving them into a litany of scalars is forcing them in the baby Perl tier; 

I don't think you understand the proposal here.

happy-barney: > I don't think you understand the proposal here.

suggesting compromise: hide this behind feature (on call side). When feature is disabled, "forced named" parameters became positional. Then it's up to user to decide.

leonerd: Yes - as @Grinnz points out, this sort of pattern is already extremely common.

It is not the point of this PPC to try to invent some fundamentally new behaviour or feature. It simply attempts to gather up an already-prevalent behaviour and formalise it into a neater way of expression. You can take justabout any CPAN module and likely find somewhere that some code unpacks its arguments into a hash, just to extract out a few fixed named keys into variables. Any of those cases could be rewritten with this feature, gaining the safety of typo protection and shorter, neater code that makes the intent easier to see because of less low-level machinery noise.

oodler577: For t

> > Then this whole PPC is pointless. When anyone is passing in named parameters with `=>`, shoving them into a litany of scalars is forcing them in the baby Perl tier;
> 
> I don't think you understand the proposal here.

Certainly possible, but show some real examples before and after where this would be useful. To most Perl developers this looks like more unnecessary complication, particularly when such an established formalism as presented in Getopt::Long is more than well suited to describe complex named parameters that do not depend on position. It also supports through various means shoving it into single scalars or into a HASH reference. And if you thought it through a bit, you could even make the resulting container class with builtin accessors.

Regarding Getopt::Long's established patterns, this would be immediately be useful for writing modulinos that need to deal with both commandline arguments and the differentiation of `caller` context.

I'd like to see some _common_ patterns, including standard idiomatic unpacking and how this would make it better.

Additinally, this seems like it could wreak havoc on any work that's been done so far on multidispatch - which unlike this PPC - seems to be actually useful to the regular Perl developer.

In summary - we need more examples so it can be evaluated by regular folks. The language would develop a lot faster and more in service to the daily Perl jocky/involuntary code maintainer if there was some separation between language design and implementation. This is just another of example of how inefficient this process is. The person who does most of the "new" development should not also be the one proposing "new" ideas that sure have been developed in a vacuum.

leonerd: Perhaps you should read the `EXAMPLES` section of this very document, to find some examples to demonstrate it.

tm604: @oodler577 For reference, here's a query to show some of the standard idiomatic examples from CPAN:

https://grep.metacpan.org/search?q=my+%5CS%2B+%3D+delete+%5C%24args%5C%7B&qd=&qft=&qifl=inc%2FModule%2FInstall%2FMakefile.pm

Currently lists over 700 distributions, give it some time to work through the rest of the search index and I'd expect that number to grow. This would seem to qualify as a _common_ pattern.

In each case, multiple lines condense into a single signature without the repetition of the name and `delete $args{...}`. That's directly useful: as a regular user, I'd be happy to use it. I know this because I'm _already_ using it, in fact, via the prototype [`Sublike::Extended`](https://metacpan.org/pod/Sublike::Extended) module mentioned in this very document. It is quite nice, and has already cut down on some finger and eyeball wear and tear, similar to the way `x // y` simplified `defined(x) ? x : y`.

As for implementing `Getopt::Long` in the signature parser, that seems like a wonderfully strange suggestion to counter "unnecessary complication", albeit not a _useful_ suggestion absent any examples or specifics. Lamentations about efficiency of process or separation of "designer" and "implementor" best directed to p5p or Facebook, though.

EvanCarroll: Hashes have overhead on access and no inherent advantage with storage. A hash ref was never designed to "organize" or namespace variables or pass arguments to a sub. It's a workaround and it comes at a cost (you're using a datastructure to workaround language semantics).

That said, this isn't without drawback. A function that takes a hash ref will have flatten the hashref it takes to pass the arguments to the function requiring named params.

```perl
sub f1 ( $args )  {
  f2( %$args );
}

sub f2 ( $x, $y, $z ) {
}
```

It was said earlier,

> The whole point as the document states, is that the **callsites for these functions do not need to be changed or know about it.**

That's only kinda-true. An advanced Perl user will only pass around hashrefs because

* it's faster than passing around scalars
* it can save the listify-hashify dance on each call.

This convention will push people into writing subs with expect lists which will make the above convention, the status quo, awkward. The real question here is whether or not the added value is worth it.

rabbiveesh: I think if we pursue destructuring assignments as mentioned in a different comment thread then the hashref vs flat list question is less problematic, bc you can unpack the kwargs hashref in one line w the same protections that signatures would provide.
Would make it one step harder for tooling like perlnavigator to help with, but I think that can be dealt with without too much trouble.

I'm a little averse to adding in a syntax to unpack kwargs from a hashref to signatures, so I would also say flat hashes ftw.

tm604: Advanced Perl users might use hashref parameters in some places, and that's not something that's going away, but I don't think the hashref/hash performance difference is that clear-cut.

Hashrefs can be a bit _slower_ when you're constructing a new/modified set of parameters - which is pretty common, compared to subs/methods that pass exactly the same parameters on to something else? It's also all too easy for something you call to remove/change something in the hashref that affects the rest of your sub:

```perl
 sub f ($args) {
  say $args->{something};
  g($args);
  say $args->{something}; # it was there just a moment ago :/
 }
 sub g ($args) {
  my $value = delete $args->{something};
 }
```

Unless you're only calling your own code, or what you're calling documents strict guarantees on not tampering with the hashref, I find that an "only-ever-hashref" API policy ends up causing more harm than good :smile: 

A simple benchmark for context here:

```
$ perl -e'
 use 5.040;
 use Benchmark qw(cmpthese :hireswallclock);
 use Sublike::Extended;
 extended sub manual (%args) {
  my $x = delete $args{x} or die "need x";
  my $y = delete $args{y} or die "need y";
  my $z = delete $args{z} or die "need z";
  die "leftover parameters" if %args;
  $x + $y + $z
 }
 extended sub named (:$x, :$y, :$z) { $x + $y + $z }
 extended sub hashref ($args) { $args->{x} + $args->{y} + $args->{z} }
 extended sub hash (%args) { $args{x} + $args{y} + $args{z} }
 cmpthese -2, {
  manual => sub { die unless 12 == manual(x => 3, y => 4, z => 5) },
  named => sub { die unless 12 == named(x => 3, y => 4, z => 5) },
  hashref => sub { die unless 12 == hashref({ x=> 3, y => 4, z => 5 }) },
  hash => sub { die unless 12 == hash(x => 3, y => 4, z => 5) }
 }
'
             Rate  manual   named hashref    hash
manual  4128545/s      --     -1%    -14%    -26%
named   4150703/s      1%      --    -14%    -26%
hashref 4812082/s     17%     16%      --    -14%
hash    5614095/s     36%     35%     17%      --
```

oodler577: > Perhaps you should read the `EXAMPLES` section of this very document, to find some examples to demonstrate it.

Thank you. The examples were the ones I saw from the PR, but more clear in https://github.com/leonerd/Perl-PPCs/blob/signature-named-params/ppcs/ppcTODO-signature-named-parameters.md.

They are still all just straight assignments for a hash key-value to a scalar. Your examples are not idiomatic and I believe contrived or cherry-picked.

It seems this is useful for a very limited sub-case where you have just a few named parameters and thus assigning key values to singular scalars is useful; like I said above, automatically replacing a litany of `my $foo = $arg{foo}` is not useful. Yeah you remove many lines of boiler plate, but that kind of pattern is not idiomatic at all anyway; hence the idiom of 

```
...
sub foo {
  my ($self, %args) = @_;
  # put some validation here if you want
}

...

my $whatever = $classic_obj_Foo->foo(arg1 => "val1", arg2 => \@val2, ... argN => $valN);
```

One non-idiomatic thing I do because I often don't have the wherewithal to look up how to do it is taking `@_` and putting it directly into a HASH ref.

What would be actually useful,

```
sub say_foo (some signature defining DSL thing ) {
  my ($self, $args_ref) = @_;
  say $args_ref->{foo};
  ...
}

$classic_obj_Foo->say_foo(foo => "hello, world!");
```

Or if you really wanted to win some internet points, add a getter:

```
sub say_foo (some signature defining DSL thing ) {
  my ($self, $args_ref) = @_;
  say $args_ref->foo;
  ...
}

$classic_obj_Foo->say_foo(foo => "hello, world!");
```

And yes, @tm604, considering the benefits of what Getopt::Long does to define a "standard" (that most of us probably know pretty well, I assume) would be great. It's familiar and battle tested. It is also supremely helpful in the use case of a modulino, as I mentioned above. Having subroutines that service a modulino that cares about `@ARGS` is very useful for testing, it it also avoids a translation layer for `@ARGS` if you're writing unit tests for said modulino.

Can you address the implications here for Corinna? I have to believe this thing that appears unhelpful in this isolated "normal Perl code" presentation is going to yield benefits down the line there? It has been pointed out elsewhere that this is more useful for the basis of "type" validation than anything else.

oodler577: > performance

There is a time and place for performance considerations, this is about language semantics and idiomatic quality of life improvements. So this is not one of those times.

perigrin: > Your examples are not idiomatic and I believe contrived or cherry-picked.

Just as a data point here, I performed the following on my current job's work repo:
```
ag '\$\w+ = \$?\$args' | wc -l
20441
```
I've learned very quickly the last couple of months that just because _I_ don't use a idiom doesn't mean that idiom isn't popular.

Basically I'm saying that if this is cherry-picked there is evidence that it's a pretty damn big cherry to contrive.

oodler577: So people who do that are going to figure out how to define a complex signature DSL statement, people who otherwise were okay with straight assignments? I mean the medicine is way worse than the cure of any potential issue. I stand of the opinion that this proposal does not stand on its own. I am looking for some situations where it does. Is it just an incomplete PPC? That's fair if it is. Just make it complete and maybe the cost of this added complexity will be more apparent to me.

oodler577:
This is my final comment and a simple summary of the insufficiency of this, which I suspect could be easy to solve.

Given,

```
sub make_colour ( :$red = 0, :$green = 0, :$blue = 0 ) { ... }

make_colour( red => 1.0, blue => 0.5 );
# The body of the function will be invoked with
#   $red   = 1.0
#   $green = 0
#   $blue  = 0.5
```

The calling pattern that this illustrates suggests one that is already common now for named parameters. However, what this gets turned into inside of the sum is not at all what is the common expected idiom. Given the `sub` interface it creates, it is reasonable to expect that there is some way to automatically achieve the equivalent of the following:

```
sub make_colour {
  my %colors = @_;
  ...
}
```

I don't know how to suggest you get both and I am not going to; however it would introduce a very confusing inconsistency between the invocation of the function and the caller's interface. This may be mutually exclusive of the desire to have a list of SCALAR variables already present.

I believe this MUST be rectified before seriously moving forward. Hopefully it's easy.

To be clear using the example currently in the PPC, this means:

```
sub make_colour ( :$red = 0, :$green = 0, :$blue = 0 ) { ... }

make_colour( red => 1.0, blue => 0.5 );
# The body of the function will be invoked with
#   $params{red}   = 1.0
#  $params{green} = 0
#   $params{blue}  = 0.5
```

Or like I said in another comment, this would be actually be the ideal (for me, anyway):

```
sub make_colour ( :$red = 0, :$green = 0, :$blue = 0 ) {}

make_colour( red => 1.0, blue => 0.5 );
# The body of the function will be invoked with 
# $_->red          a "reader" or getter that returns 1.0, with direct access to $_->{red}
# $_->green       a "reader" or getter that returns 0, with direct access to $_->{green}
# $_->blue        a "reader" or getter that returns 0.5, with direct access to $_->{blue}
```

Not sure what to say about the package string or `$self` if it's called that way; but that ambiguity seems present in the current spec also.

If you really want to have `$red`, `$blue`, and `$green` available then do it. But having now way to easily capture the kv pairs easily into a offers no improvement to the situation where one may do, `my %params=@_` - where as providing that HASH ref would benefit that in at least a couple of ways according to my thinking (`$_` is a HASH ref, and if getters are provided it makes thing nice and clean for non-`class` Perl).

I have no opinion on the syntax DSL defining the signature, there's only so much you can do there. Hopefully something I have said or suggested above will result in something positive.

ppcs/ppcTODO-signature-named-parameters.md
tonycoz: While ugly, if we don't want to support some special syntax for this (I think) rare cae, the function could set the default to an unlikely value and check for that:
```
my $unknown;
# x can be supplied as any value, but \$unknown isn't accessible elsewhere
sub f ( :$x = \$unknown) { 
  if (ref $x && $x == \$unknown) {
    # whatever we want to do that's special
  }
}
```

tonycoz, line 147: Another thing to consider here is multiple names for the same parameter, eg, your make_colour() might accept `red`, `green` and `blue` or `r`, `g` and `b`.

zmughal:
Raku equivalent for comparision:
- [Synopsis 6: Subroutines](https://github.com/Raku/old-design-docs/blob/master/S06-routines.pod#named-parameters)
- [roast test suite](https://github.com/Raku/roast/tree/@{2024-08-15}/S06-signature)

ppcs/ppcTODO-signature-named-parameters.md
zmughal: > Another thing to consider here is multiple names for the same parameter, eg, your make_colour() might accept `red`, `green` and `blue` or `r`, `g` and `b`.

I know that in PDL and R (e.g., `ggplot2`), it's typical to normalise named parameters to accept `colour` and `color`. Though I can see this kind of feature being handled via an attribute (or keep the custom code that is already there).

Raku's synopsis that I linked in a comment below gives the example of:

```raku
sub globalize (:g(:global($gl))) {...}
```

where the variable is `$gl`, but the named key is either `global` or `g`.

ppcs/ppcTODO-signature-named-parameters.md
zmughal: This kind of thing would be better handled via `multisub`s, no? That's kind of what `Type::Params` [allows](https://metacpan.org/release/TOBYINK/Type-Tiny-2.004000/view/lib/Type/Params.pm#multiple-ArrayRef), but I don't think that is good to have in a general case.

rabbiveesh:
> Given the `sub` interface it creates, it is reasonable to expect that there is some way to automatically achieve the equivalent of the following:
> 
> ```
> sub make_colour {
>   my %colors = @_;
>   ...
> }
> ```
> 
> I don't know how to suggest you get both 


If you'd like that interface you could have it now, using `sub make_color (%colors) {}`. Are you sure you know perl?

ppcs/ppcTODO-signature-named-parameters.md
zmughal: Here's a possible approach:

```perl
sub f(
    # Internally as C<$colour> with external API of I<colour> and I<color>.
    :$colour :name(color),

    # Internally as C<$gl> with external API of I<global> and I<g>.
    # NOTE that C<$gl> does not have a colon (C<:>) preceding it.
    # The :name() indicates that it is not positional.
    $gl      :name(global g),

) {
    say "Color: $colour, Global: $gl";
}

sub c(
    :$red   :name(r), # both red   and r
    :$green :name(g), # both green and g
    :$blue  :name(b), # both blue  and b
) {
    ...
}
```

This aliasing approach does not have to be examined within this PPC and the
contents of this PPC don't conflict with it.

happy-barney: @zmughal you are right this goes further then this PPC, there are tons of things ...

it depends on what we want to do with Perl itself. Focus on readability? Focus on catching train? Focus on defining strong side where other languages cannot compete yet? Focus in expressing implementation? Focus on expressing intention? 

Regarding readability: I suggest to read this: https://www.scientificamerican.com/blog/frontiers-for-young-minds/the-reading-brain/

there are also studies related to readability of code - though every such article/study will get opposition in style "I'm against this because I can read my code I just wrote"

So regarding this PPC:
- will it bring short-time improvement? yes
- will it bring long-time improvement? syntax-wise I'm skeptical, but I think it will bring enough improvements in core to make better syntax improvement possible

What this PPC is missing (according me) is comparison how other languages deal with problem and evaluation of pros/cons of such solution. At least first part I did some time ago: https://github.com/happy-barney/perl-wish-list/blob/master/named-arguments/spec/named-arguments-in-other-languages.md

iabyn (commented):
Looking at this proposal compared to my earlier 2019 signatures proposal
(http://nntp.perl.org/group/perl.perl5.porters/256679):

My proposal allowed *optional* positional parameters as long as all named parameters were also optional:

```
    sub f($x, $y = 0, :$r = 0, :$g = 0, :$b = 0);
    f(1);
```

My proposal insisted that all mandatory named params come before optional named params.

More generally, the processing order of default value expressions and fetching of arg values (e.g. FETCH calls on tied values) needs to be carefully considered. So for example in:

```
    sub f(:$a = def_a(), :$b = def_b(), :$c = def_c())
    f(c => $tied_c, a => $tied_a)
```

in which order are $tied_a->FETCH(), def_b(),  $tied_c->FETCH() called? We either need to document the order, or explicitly document that the order is undefined and subject to change, because otherwise any later internal optimisations will break someone's code.

My original proposal suggested that there should be a conceptual sorting and de-duplicating of the name value pairs in the args list into the same order as the declared params, *then* arg to param binding takes place in L-R order. How this is implemented internally is up to us, but as long as the visible effects of ordering is preserved, it doesn't matter.

Only allowing a slurpy hash: why not allow slurpy array?

Duplicate named args *shouldn't* warn by default, to allow caller to do
`
    foo(%defaults, %args)
`
for example. I also suggested that for duplicate args, only the second would have its value accessed (i.e. get magic called, e.g. FETCH).

dstroma:
How would collisions with Corinna class field names be handled?
```
class Point {
  field $x;
  field $y;
  method add (:$x, :$y) {
    # Whose $x is $x?
```

Or will it cause an error?

Grinnz:
> How would collisions with Corinna class field names be handled?
> 
> ```
> class Point {
>   field $x;
>   field $y;
>   method add (:$x, :$y) {
>     # Whose $x is $x?
> ```
> 
> Or will it cause an error?

I would suggest the same thing would happen as if you do `field $x; { my $x; }` since this is effectively a lexical variable declaration.

oodler577:
After thinking more, I didn't handle any of this the right way. i am sorry for any problems this caused, I should have done my thinking and processing privately. Now having done so, I am going to try a different tact.

**_Please add to this a way to access a HASH reference that contains the respective named parameters as keys that point to the passed values._** I assume this would leverage references so that `$all_params->{red}` was a reference to `$red`.  Keep the automagical scalars, e.g., `$red`, etc.

This simple addition would make this feature extremely useful for Perl programmers who use named parameters often and already do the `my %params = @_` unpacking, but in this case it is a value add because as a reference, `$params` (or whatever you name it, idc), would already be unpacked and already to go.

The automagical named scalars are useful for beginners and for short subroutines by everyone (I admit, this could be useful in some cases); but for more complicated subroutines and more advanced programmers, access to the params as a HASH ref (as many tend to do) will be missing and they will have to fall back to `my %params = @_` - the way they would do it anyhow, leaving them with zero reason to use the signatures in the first place.

I can't say for certain, but I imagine the overhead here is negligible; the benefits are extended to a much wider and experienced audience. Please consider this as part of your proposal here for named parameters.

Thank you.

mauke:
> The automagical named scalars are useful for beginners and for short subroutines by everyone (I admit, this could be useful in some cases);

It's not just beginners and short subs. The lack of named parameters in core is the main reason I still use [Function::Parameters](https://metacpan.org/pod/Function::Parameters), even in CPAN code. (NB: Function::Parameters has had named parameters using the exact syntax proposed above since 2012. This is hardly a new or controversial feature.)

> but for more complicated subroutines and more advanced programmers, access to the params as a HASH ref (as many tend to do) will be missing and they will have to fall back to `my %params = @_` - the way they would do it anyhow, leaving them with zero reason to use the signatures in the first place.

Some of the things I don't understand:

- Why do you think a reference to a hash is better than a hash?
- Why do you think the more complicated the subroutine, the better it is to have unchecked variable references (in the form of `$params->{foo}` or `$params{foo}`) flying around (as opposed to `$foo`, which is checked at compile time)?
- If `my %params = @_` does what you want, then there is no reason for you to use "signatures".¹ That's fine. (Ditto for `my $params = {@_}` and hash references.) How is this a problem?
- That said, `sub foo(%params)` would give you arbitrary key/value pairs under the current proposal. What's wrong with just using that?

---

¹ I still hate that name. They're just parameter lists.

happy-barney:
> > The automagical named scalars are useful for beginners and for short subroutines by everyone (I admit, this could be useful in some cases);
> 
> It's not just beginners and short subs. 

Generally named arguments improves readability and, when well named, intention of author. It lowers learning costs of newbies in project as well as can be difference between maintainable software and legacy piece of mud.

oodler577:
> Why do you think a reference to a hash is better than a hash?

I don't, a hash is fine.

> That said, `sub foo(%params)` would give you arbitrary key/value pairs under the current proposal. What's wrong with just using that?

The information communicated with the explicitly defined params in the signature gives a very helpful level of self documentation about the subroutine using it. The value of the named parameters signature to me is clear, this is building on top of that.

> Generally named arguments improves readability and, when well named, intention of author. It lowers learning costs of newbies in project as well as can be difference between maintainable software and legacy piece of mud.

I agree, I see their benefit. Providing key/pair access to the passed parameters would make this PPC even more helpful by providing advanced idiomatic access to the arguments passed into the subroutine. It also takes care of providing introspection inside of the subroutine about what parameters are available via the signature. As far as I can tell, this information would have to be held externally if one wanted to leverage that list.

Thank you.

happy-barney:
> I agree, I see their benefit. Providing key/pair access to the passed parameters would make this PPC even more helpful by providing advanced idiomatic access to the arguments passed into the subroutine. It also takes care of providing introspection inside of the subroutine about what parameters are available via the signature. As far as I can tell, this information would have to be held externally if one wanted to leverage that list.
> 

well, `%_` can be abused to do that (though there I'd prefer keys with sigil)

esabol:
@mauke wrote:
> The lack of named parameters in core is the main reason I still use [Function::Parameters](https://metacpan.org/pod/Function::Parameters), even in CPAN code. (NB: Function::Parameters has had named parameters using the exact syntax proposed above since 2012. This is hardly a new or controversial feature.)

Thank you. This provided some very useful context for this discussion, I feel.

adamcrussell:
> As is traditional for me I didn't fill in the "Rationale" section as I still have no idea how that differs from a "Motivation". The two words seem rather synonymous to me.

Rationale would seem to be the more abstract purpose of the proposal and Motivation an example of an actual problem that would be solved or a situation that would be made easier.

To that end, do you have a practical minimal example of the problem or situation that this would address? Perhaps, say, you encountered something in a recent application or project you are developing outside of your work on the interpreter?

zmughal:
I wrote up some code that has several approaches to using named parameters in both the explicit *scalar-per-parameter* approach and *slurpy-hash* approach here <https://github.com/zmughal-experiment/perl-parsing-panoply/pull/1>. My goals were to:

- see how well it works with `Type::Tiny` checks (*slurpy-hash* that is handled without any objects would need checks over the entire container and coercion to support default values for optional elements  --- in lieu of currently non-existent destructuring syntax that might be able to support default values);
- see if there was some way of making the parameters introspectable (to see how DRY one can get by using lists from either the parameter list, `Type::Tiny` object, or a `MooX::Struct`-based parameter object);
- reusability of a signature across multiple functions (i.e., for defining a common interface).

I can see using both (`xor`) approaches on a case-by-case basis, but wanting to still have the same external API. Perhaps this can be done via the `:collect` idea, but instead the attribute can also specify a variable to collect into (e.g.,
```
sub f( :$red :collect(%color), :$blue :collect(%color), ... ) { ... }
```
or an attribute on the entire function
```
sub f :collect(@pos %named) ($one, $two, :$r, :$g, :$b) ) {
  $params{r}
}
```
)

The particular example of using an RGB color triple might not be the best way to compare as that points towards using an object as these are closely related values.

---

Another thing that I would like to think about when it comes to working with parameters is how easy is it to make sure that wrapper functions and functional programming are easy to write (e.g., an `around` modifier that modifies or splices out a single positional or named parameter and then sends the argument list on to the coderef that it is wrapping). I previously mentioned that [here](https://github.com/orgs/Perl-Apollo/discussions/43#discussioncomment-6815368) in relation to how checks in the signature might work.

oodler577:
> > I agree, I see their benefit. Providing key/pair access to the passed parameters would make this PPC even more helpful by providing advanced idiomatic access to the arguments passed into the subroutine. It also takes care of providing introspection inside of the subroutine about what parameters are available via the signature. As far as I can tell, this information would have to be held externally if one wanted to leverage that list.
> 
> well, `%_` can be abused to do that (though there I'd prefer keys with sigil)

Thank you for the chance to clarify. I hope some kind of hash access will be considered.

ppcs/ppcTODO-signature-named-parameters.md
haxmeister: named parameters appeal to me but adding the :$param syntax in the signature section feels heavy on my mind when considering how feature-rich our subroutines already are. Might I offer a possibility to help you avoid the : and distinguish it from the other style arguments? Perhaps an optional block that contains named parameters?
```perl
sub make_color ( $arg1, @slurpy) < $red = "red",  $blue = "blue"> {
     say "sub's body";
}
```
they < > wrappers may not be the best choice.. but the point is that a different/separate wrapper provides a context for less contrived syntax in the named var assignments and/or confusion about how params work in perl to newcomers

leonerd:
@iabyn 
> Looking at this proposal compared to my earlier 2019 signatures proposal (http://nntp.perl.org/group/perl.perl5.porters/256679):
> 
> My proposal allowed _optional_ positional parameters as long as all named parameters were also optional:
> 
> ```
>     sub f($x, $y = 0, :$r = 0, :$g = 0, :$b = 0);
>     f(1);
> ```

Yes - your idea, plus also there's a few other CPAN modules (Function::Parameters, Kavorka, etc..) all used the same idea, which is why I went with that.

I don't think I've explicitly put words in to this effect, but yes - if all the named ones are optional then there can be optional positional ones too. I can add some words for that.

> My proposal insisted that all mandatory named params come before optional named params.

Mmm, that restriction might become necessary depending on the implementation, but the way I originally implemented it in XS::Parse::Sublike didn't require it, so I didn't write about that. Perhaps we can leave it flexible for now awaiting an actual implementation.

> More generally, the processing order of default value expressions and fetching of arg values (e.g. FETCH calls on tied values) needs to be carefully considered. So for example in:
> 
> ```
>     sub f(:$a = def_a(), :$b = def_b(), :$c = def_c())
>     f(c => $tied_c, a => $tied_a)
> ```
> 
> in which order are $tied_a->FETCH(), def_b(), $tied_c->FETCH() called? We either need to document the order, or explicitly document that the order is undefined and subject to change, because otherwise any later internal optimisations will break someone's code.

Ahyes a good idea .My original plan was to leave details like that relatively unspecified in the PPC doc and defer such considerations to the actual implementation and its documentation. Perhaps I should at least write words in the PPC to explain that bit.

> My original proposal suggested that there should be a conceptual sorting and de-duplicating of the name value pairs in the args list into the same order as the declared params, _then_ arg to param binding takes place in L-R order. How this is implemented internally is up to us, but as long as the visible effects of ordering is preserved, it doesn't matter.

Yes, more ordering things to consider.

> Only allowing a slurpy hash: why not allow slurpy array?

That fell out of the way I implemented it in XS::Parse::Sublike. What XPS does is creates a lexical hash with an unparseable name to unpack the caller args into and process them into the individual vars. If the signature syntax wanted a slurpy hash at the end, well it became a simple matter to just use *that* name as the internal hash name anyway, then it's visible to the subroutine body for free. I couldn't think of a way to implement that with an array, without just assigning back out of the hash into an array and possibly breaking the originally-passed values.

If we allow a slurpy array, it leads to questions about whether it just has list-like nature of caller-passed values that don't follow the unique key/value structure of a hash. For example:

```perl
sub f($alpha, $beta, :$x, :$y, @rest) {
  say for @rest;
}

f(12, 34, x => 5, y => 6, z => 7, z => 8, z => 9);
```

In this case, would `@rest` see a 6-value list `('z', 7, 'z', 8, 'z', 9)`? How does that interact with other orders of values being passed by the caller?

> Duplicate named args _shouldn't_ warn by default, to allow caller to do `foo(%defaults, %args)` for example. I also suggested that for duplicate args, only the second would have its value accessed (i.e. get magic called, e.g. FETCH).

Yup, sounds sensible.

ppcs/ppcTODO-signature-named-parameters.md
leonerd: @zmughal 
> Here's a possible approach:
> 
> ```perl
> sub f(
>     # Internally as C<$colour> with external API of I<colour> and I<color>.
>     :$colour :name(color),

Yes, an attribute on the parameter would seem to be a sensible attack at that. We don't currently have an ability to put attributes in positions like that, but `XS::Parse::Sublike` permits that sort of thing. It's the subject of a different PPC of mine that I'm still in the middle of drafting. Perhaps I can extend the wording of this one to at least hint at it though.

leonerd:
Following up on some comments here, I've added some more wording. I've specifically mentioned a few other CPAN modules that provide the same syntax idea, and expanded on the suggestion of an attribute to provide alternative names for parameters.

iabyn:
On Fri, Aug 23, 2024 at 03:11:54AM -0700, Paul Evans wrote:
> Mmm, that restriction might become necessary depending on the implementation, but the way I originally implemented it in XS::Parse::Sublike didn't require it, so I didn't write about that. Perhaps we can leave it flexible for now awaiting an actual implementation.

In general I would prefer we restrict things slightly *now* - at least for
things which don't inconvenience the coder - in order to allow more
flexibility for future optimisations. This comes from long painful
experience of doing optimisation work over the years. For example,
multiconcat slightly changes the order in which FETCH() methods etc are
called on concat arguments (IIRC), and this broke some CPAN modules.

> > Only allowing a slurpy hash: why not allow slurpy array?
> 
> That fell out of the way I implemented it in XS::Parse::Sublike. What XPS does is creates a lexical hash with an unparseable name to unpack the caller args into and process them into the individual vars. If the signature syntax wanted a slurpy hash at the end, well it became a simple matter to just use *that* name as the internal hash name anyway, then it's visible to the subroutine body for free. I couldn't think of a way to implement that with an array, without just assigning back out of the hash into an array and possibly breaking the originally-passed values.
> 
> If we allow a slurpy array, it leads to questions about whether it just has list-like nature of caller-passed values that don't follow the unique key/value structure of a hash. For example:
> 
> ```perl
> sub f($alpha, $beta, :$x, :$y, @rest) {
>   say for @rest;
> }
> 
> f(12, 34, x => 5, y => 6, z => 7, z => 8, z => 9);
> ```
> 
> In this case, would ***@***.***` see a 6-value list `('z', 7, 'z', 8, 'z', 9)`? How does that interact with other orders of values being passed by the caller?

I think I addressed that in my original proposals.

NB: my vague intention for implementing named subs was an algorithm along
the lines of:

At compile time, construct some sort of lightweight perfect hash
(*not* a perl hash, at least not eventually) that maps every legal
name key to an index number. This hash would be attached to the
OP_ARGCHECK.

At runtime, pp_argcheck() would extend the stack by 2 x (number of named
parameters), and fill with NULLs. Then the args (starting at the
position after all fixed parameters and going L-R) would be grabbed,
with each (name, value) pair inserted at the index looked up based on
the name. This is effectively a pigeon-hole stable sort with
de-duplication.

Then subsequent arg processing ops would run along the extended stack
L-R, assigning the values to the pad vars in order, and for any slots
still NULL, call the associated default value expression. The
mechanism can treat them more-or-less like positional parameters now.

This is also based on the assumption that the original args will
eventually be on the stack rather than in @_.

That's very vague, and there's a lot more to it that that, but one thing
I've always been very conscious of is that perl subroutine calls are *very*
slow - they have a lot of overhead. We should be designing any signature
features in such a way that they're no slower than currently, and in a way
which provide the long-term potential for big speedups. I would consider
populating a perl hash for each subroutine call a heavy overhead.

ilmari:
> @iabyn
> 
> > Looking at this proposal compared to my earlier 2019 signatures proposal (http://nntp.perl.org/group/perl.perl5.porters/256679):
> > My proposal allowed _optional_ positional parameters as long as all named parameters were also optional:
> > ```
> >     sub f($x, $y = 0, :$r = 0, :$g = 0, :$b = 0);
> >     f(1);
> > ```

How would this work if there are two optional positional parameter, and you want to pass a value to the first one that happens to equal a named parameter?

```
sub f($x = '', $y = 0, :$z = 0);
f(z => 42);
```
or more subtly:
```
my ($foo, $bar) = qw(z 42);
f($foo, $bar);
```

leonerd:
@ilmari 
I would imagine if you wanted to pass any named parameters, then you'd have to fill in any of the optional ones first. Which is mostly why I didn't really write about it or consider it in my thoughts. Most of the point of having named parameters is to allow you to have optional params without specifying a "priority order" to which ones you have to fill in before you can pass the others. We might specify that "hey, you can do this, but it probably won't do anything you find useful, so maybe don't do that?"

dstroma:
> > How would collisions with Corinna class field names be handled?
> > ```
> > class Point {
> >   field $x;
> >   field $y;
> >   method add (:$x, :$y) {
> >     # Whose $x is $x?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or will it cause an error?
> 
> I would suggest the same thing would happen as if you do `field $x; { my $x; }` since this is effectively a lexical variable declaration.

Indeed, you can already shadow a field variable with a positional parameter, so doing it with a named parameter isn't really anything new. Might be worth it to consider a warning if this happens but that is beyond the scope of this PPC.

I like this PPC as is. I'd rather not overcomplicate it by allowing such things as multiple names for the same parameter. Those who want that functionality can just use the traditional way of doing that, or a CPAN module. Or it can be a future PPC.

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,185 @@
...
+sub f {
+    die "Too many arguments" if @_ > 3;
+    die "Too few arguments" if @_ < 2;
+    my ($x, $y, $z) = @_;
+    $z = 123 if @_ < 3;
+    ...
+}
+```
+
+One common form of argument processing involves passing an even-sized list of key/value pairs and assigning that into a hash within the subroutine, so that specifically named parameters can be extracted from it. There is currently no support from subroutine signature syntax to assist authors in providing such behaviours.
```
zmughal, line 31: An example of where this could be used is command-line argument processing. Raku makes the [`MAIN` subroutine](https://docs.raku.org/language/create-cli) special and uses the signature to give a default `getopt`-style interface.

leonerd:
@dstroma 
> I like this PPC as is. I'd rather not overcomplicate it by allowing such things as multiple names for the same parameter. Those who want that functionality can just use the traditional way of doing that, or a CPAN module. Or it can be a future PPC.

Indeed - this is the point of the "Future Scope" and "Open Issues" parts of the document. They're for explaining "here's some things we thought about but we're not solving at the moment".

iabyn:
On Fri, Aug 23, 2024 at 05:44:56AM -0700, Dagfinn Ilmari Mannsåker wrote:
> How would this work if there are two optional positional parameter, and you want to pass a value to the first one that happens to equal a named parameter?
> 
> ```
> sub f($x = '', $y = 0, :$z = 0);
> f(z => 42);

Positional parameters consume any args first. Only if there are any args
left after all positional params have been satisfied, should named param
processing start. So your example above behaves equivalently to:
```
  my $x = 'z';
  my $y = '42';
  my $z = 0;   # default value
```

-- 
"But Sidley Park is already a picture, and a most amiable picture too.
The slopes are green and gentle. The trees are companionably grouped at
intervals that show them to advantage. The rill is a serpentine ribbon
unwound from the lake peaceably contained by meadows on which the right
amount of sheep are tastefully arranged." -- Lady Croom, "Arcadia"

leonerd:
I've added further wording on some more comments.

One thing I've not yet addressed is the relative order between checking for the presence of mandatory named parameters, vs evaluating any defaulting expressions. With positional parameters it's easy to make a check of `scalar @_` between numerical limits, to see whether this implies the minimum number of required positional parameters. This check is done early (as part of the `OP_ARGCHECK` opcode), meaning we reject invalid calls that are missing arguments even before any defaulting expressions are evaluated.

But with named ones we'd have to scan over the entire incoming arguments list to look for what names are present. In the current implementation I have made in `XS::Parse::Sublike`, the check for mandatory named parameters is effectively part of evaluating the defaulting expression, because I just store a defaulting expression that throws an exception complaining of a missing argument. In effect that means the check happens as a side-effect of evaluating the default value, but means that each parameter is only checked for presence *after* any earlier parameters have already had their defaulting expressions checked.

In `XS::Parse::Sublike` I specifically document that this order is not guaranteed, to give some flexibility here, but it's likely something that a core implementation should think more about, and maybe find a better solution to.

ppcs/ppcTODO-signature-named-parameters.md
zmughal: This is also out of scope and for more general use, but what about the use of the experimental `refaliasing`/`declared_refs` feature when it comes to use with parameter syntax? I know that @leonerd has <https://metacpan.org/pod/Signature::Attribute::Alias> which does aliasing via a parameter attribute syntax.

ppcs/ppcTODO-signature-named-parameters.md
zmughal: @haxmeister The proposal is not that far off from other languages. Raku uses a similar syntax. The colon prefix is similar to the convention of using a colon for keyword symbols in Common Lisp (though this is at the call site).

Also, all languages I am familiar with that have named parameters allow for a mix of positional and named parameters in a single parameter list in their signatures, so that shouldn't be novel.

happy-barney: > @haxmeister The proposal is not that far off from other languages. Raku uses a similar syntax. The colon prefix is similar to the convention of using a colon for keyword symbols in Common Lisp (though this is at the call site).
> 
> Also, all languages I am familiar with that have named parameters allow for a mix of positional and named parameters in a single parameter list in their signatures, so that shouldn't be novel.

Just to point out, most of languages doesn't have sigils and treat list variables as scalars, so not providing sigil in call syntax is designed to lose information.

haxmeister: The syntax by itself is not the reason that I found this to be heavy on my mind. Rather it it the syntax in light of how it is used in the current class feature. I have already associated the  colon as an attribute syntax that follows a convention where the colon is followed by the name of a behavior that is applied to the preceding var. This proposed syntax will apply the colon directly to the var.. it is not the same thing and my mind immediately rejected using it this way after having written much code using the new class syntax.

zmughal: @haxmeister, yeah, I can see that. I like the extensibility of the KIM syntax that was mentioned before in the design of the `class` feature.

But choosing a single ASCII character not used elsewhere in the Perl syntax is also difficult. Named parameters are probably going to be a commonly used feature so shorter is better.

Perhaps looking at how Kavorka does it <https://metacpan.org/dist/Kavorka/view/lib/Kavorka/Manual/Signatures.pod#Named-parameters>, <https://metacpan.org/dist/Kavorka/view/lib/Kavorka/Manual/Signatures.pod#Long-name-parameters> which is based on the form used in Raku, might help.

If the approach that @leonerd has of having attributes in the parameter list goes through, I think using a longer form with a `:name()` parameter attribute that repeats the same name as the variable could be a pattern to make the syntax stand out. I typically write out anything with many parameters and additional default values so that each parameter is on a line by itself.

ppcs/ppcTODO-signature-named-parameters.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Parameter Names
+
+How to specify a named parameter whose name is anything other than the name of the lexical variable into which its value is assigned? This is more of note when considering that traditional argument handling techniques involving assignment into hashes can handle more styles of name that would be invalid for lexical variables - for example, names including hyphens.
+
+Perhaps a solution would be to permit attributes on parameter variables, then define a `:name` attribute for this purpose.
+
+```perl
+sub display ( $message, :$no_colour :name(no-colour) ) {
+    ...
```
zmughal, line 181: The semantics of this aren't defined here, but if this acts as simply an override on the parameter key name rather than an addition to a list of parameter key names, I would like that over my earlier suggestion of not having a colon. :-)

leonerd:
Based on this comment from the PSC, I've given the document the official 0024 numbering.
https://www.nntp.perl.org/group/perl.perl5.porters/2024/09/msg268790.html

I guess this is ready to merge now?

PR #56, opened 2024-10-07
New PPC: Attributes v2

leonerd:
## Abstract

Define a new way that attribute definitions can be introduced such that the parser can invoke the third-party custom logic they provide; and additionally permit them to be attached to more types of target than Perl currently allows.

---

Still currently in draft as I work out a few more minor questions, but now at the stage where it should be suitable for comments.

### DIFF
--- a/ppcs/ppcTODO-attributes-v2.md
+++ b/ppcs/ppcTODO-attributes-v2.md
@@ -29,9 +29,9 @@ The parser supports additional attributes that can be provided by modules, thoug
 
 * Only lexical or package variables, and subroutines, support attribute syntax. It is not possible to declare attributes on other entities such as packages, or subroutine parameters.
 
-* Attributes on anonymous subroutines are invoked only at "clonecv"-time; the time when an anonymous function gets turned into a closure with variable captures. It cannot perform any behaviour before this time.
+* Attributes on anonymous subroutines are invoked only once at compiletime on the protosub, before any "clonecv" operation that would make a real closure. It cannot perform any behaviour after this time, and does not get to see the real closure CVs that Perl code gets references to.
 
-* Attribute handling is looked up via the package heirarchy expressed in `@ISA`, rather than by lexical scope.
+* Attribute handling is looked up via the package hierarchy expressed in `@ISA`, rather than by lexical scope.
 
 * Third-party attributes are implemented by providing a single `MODIFY_*_ATTRIBUTES` shouty-named method in a package, which is required to understand all the attributes at once. There is no standard mechanism to create individual attributes independently.
 
@@ -49,7 +49,7 @@ Attributes defined by this specification will be lexical in scope, much like tha
 
 An attribute is defined primarily by a C callback function, to be invoked by the parser _as soon as_ it has finished parsing the declaration syntax. This callback function will be passed the target (i.e. the item to which the attribute is being attached), and the optional contents of the parentheses used as an argument to the attribute. There is no interesting return value from this callback function.
 
-An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenience to have to represent the parameters to an attribute as a flat text string.
+An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenient to have to represent the parameters to an attribute as a flat text string.
 
 Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact parameter types are still an open issue; see below):
 
@@ -58,7 +58,7 @@ struct PerlAttributeDefinition
 {
     U32 ver;
     U32 flags;
-    SV * (*parse)(pTHX_ SV *text);   /* optional, may be NULL */
+    SV * (*parse)(pTHX_ const SV *text);   /* optional, may be NULL */
     void (*apply)(pTHX_ SV *target, SV *attrvalue);
 };
 ```
@@ -139,7 +139,7 @@ The mechanism proposed here makes the following improvements over the existing a
 
 ## Rejected Ideas
 
-* Filtering or flags on attribute definitions to say what kind of target they apply to. By omitting this, a simpler model is provided. It avoids complex questions on how to handle heirarchial classifications, such as that classes are packages, or subroutine parameters are lexical variables. By ensuring that any particular attribute name has at most one definition in any scope, end-users can more easily understand the model provided.
+* Filtering or flags on attribute definitions to say what kind of target they apply to. By omitting this, a simpler model is provided. It avoids complex questions on how to handle hierarchial classifications, such as that classes are packages, or subroutine parameters are lexical variables. By ensuring that any particular attribute name has at most one definition in any scope, end-users can more easily understand the model provided.
 
 ## Open Issues
 
--- a/ppcs/ppcTODO-attributes-v2.md
+++ b/ppcs/ppcTODO-attributes-v2.md
@@ -53,7 +53,7 @@ An attribute is defined primarily by a C callback function, to be invoked by the
 
 An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenient to have to represent the parameters to an attribute as a flat text string.
 
-Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact parameter types are still an open issue; see below):
+Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact type is still an open issue; see below):
 
 ```c
 struct PerlAttributeDefinition
@@ -170,16 +170,9 @@ It may be the case that at import time, extra information is attached to the (un
 
 It would first appear that lexical variables and subroutine parameters can be represented by their PADNAME structure, but notably the padname itself does not actually store the pad offset of the named entity. Perhaps the target argument for these should just be the pad offset of the target entity, leaving the invoked callback to find the offset in the compliing pad itself?
 
-This might suggest that actually the target should be specified as two - or maybe even three - parameters, some of which would be zero / NULL depending on circumstance.
+This suggests that the actual values passed to specify the target will depend on what kind of target it is. Package-named targets can be passed the target SV itself and its naming GV, whereas lexical targets need to be specified as its pad offset within the currently-compiling pad.
 
-```c
-    void (*apply)(pTHX_
-        SV *target, GV *targetname, /* for package targets, or NULL/NULL */
-        PADOFFSET targetix,         /* for lexical targets, or zero */
-        SV *attrvalue);
-```
-
-But perhaps at that point, it makes more sense to have two different callbacks, one for GV-named targets and one for lexicals?
+It may make sense to have two different callbacks, one for GV-named targets and one for lexicals
 
 ```c
 struct PerlAttributeDefinition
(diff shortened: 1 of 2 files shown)


ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+Additionally, new callsites can be added to the parser to invoke attribute callbacks in new situations that previously were not permitted; such as package declarations or subroutine parameters.
+
+Note that this specification does not provide a mechanism by which attributes can declare what kinds of targets they are applicable to. Any particular named attribute will be attempted for _any_ kind of target that the parser finds. This is intentional, as it leads to a simpler model both for the interpreter and for human readers of the code, as it means any particular attribute name has at most one definition; its definition does not depend on the type of target to which it is applied. It is the job of the attribute callback itself to enquire what kind of target it has been invoked on, and reject it if necessary - likely with a `croak()` call of some appropriate message.
```
happy-barney, line 78: when `parse` is invoked, it knows context, so why not to defined enumeration of context identifiers and send it along with text, so it can return value aware of such context?

should this "enumeration be extendable, such mechanism can be reused by 3rd party extensions to declare new contexts for their use (along with core contexts) and use attribute handling provided by core

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+It would be a nontrivial undertaking, but it may become useful in terms of memory (and CPU cycles) savings to consider splitting the pad into two separate entities. A "scope" pad would exist just once per file or subroutine, and contain lexically-imported named elements such as imported functions or attribute definitions. Separately a "scratchpad" pad would then only contain the lexical variables and temporary values that are needed once for every depth of recursion of the function. By splitting the pad in such a way, it reduces the amount of work that `pp_entersub` has to perform when recursing into functions that have wide pads because now only the true variables-per-call need to be cloned; the static scope would exist just once for all depths.
+
+### Perl-Level API
+
+The mechanisms described in this document exist entirely at the C level, intended for XS modules to make use of. It should be possible to provide wrappings of most of the relevant parts, at least for simple cases, for use directly by Perl authors without writing C code.
```
happy-barney, line 208: I'd suggest to define reusable (from core as well from 3rd party modules) syntax to declare additional symbols (taken from my "independent symbol spaces"). For attributes:
```
declare attribute Foo as
    on CODE   => { ... do something ... }
    on SCALAR => { ... do something ... }
    on ARRAY | HASH => { ... do something ... }
  ;
```

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+(explain why the (following) proposed solution will solve it)
+
+## Specification
+
+Attributes defined by this specification will be lexical in scope, much like that of a `my` variable, `my sub` function, or any of the builtin function exports provided by the `builtin` module. This provides a clean separation of naming within the code.
+
+An attribute is defined primarily by a C callback function, to be invoked by the parser _as soon as_ it has finished parsing the declaration syntax. This callback function will be passed the target (i.e. the item to which the attribute is being attached), and the optional contents of the parentheses used as an argument to the attribute. There is no interesting return value from this callback function.
+
+An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenience to have to represent the parameters to an attribute as a flat text string.
```
mauke, line 52: \*inconvenient

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact parameter types are still an open issue; see below):
+
+```c
+struct PerlAttributeDefinition
+{
+    U32 ver;
+    U32 flags;
+    SV * (*parse)(pTHX_ SV *text);   /* optional, may be NULL */
+    void (*apply)(pTHX_ SV *target, SV *attrvalue);
+};
```
mauke, line 63: In general, a "function value" should be represented in C as a pair of (function pointer, context pointer) to enable closure-like behavior. So I would expect to see something more like
```c
    void *parse_ctx;
    SV *(*parse)(pTHX_ void *ctx, SV *text);
    void *apply_ctx;
    void (*apply)(pTHX_ void *ctx, SV *target, SV *attrvalue);
```

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenience to have to represent the parameters to an attribute as a flat text string.
+
+Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact parameter types are still an open issue; see below):
+
+```c
+struct PerlAttributeDefinition
+{
+    U32 ver;
+    U32 flags;
+    SV * (*parse)(pTHX_ SV *text);   /* optional, may be NULL */
```
mauke, line 61: What does `parse` return? If `parse` isn't supposed to modify the `text` parameter in place, shouldn't it be marked `const`?

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Perl-Level API
+
+The mechanisms described in this document exist entirely at the C level, intended for XS modules to make use of. It should be possible to provide wrappings of most of the relevant parts, at least for simple cases, for use directly by Perl authors without writing C code.
+
+Existing experience with [`XS::Parse::Keyword::FromPerl`](https://metacpan.org/pod/XS::Parse::Keyword::FromPerl) shows that while such an interface could easily be provided, in practice it is unlikely anyone would make use of it as it requires understanding details of its operation in sufficient depth that any author would be just as well served by writing the XS code directly. Therefore no such wrapping is yet proposed by this document but could be added later on, either by a core-provided or CPAN module.
```
mauke, line 210: I think for Perl-level operations it makes much more sense to have the opposite of this proposal: Attributes that apply to Perl entities (variables, subroutines), not parser/interpreter artifacts (GVs, pad slots, CVs, protosubs). Since most of these entities (lexical variables, anonymous closures) only exist at runtime, it would mean providing a runtime hook that is invoked when the entity is instantiated.

But that's a very different proposal.

ppcs/ppcTODO-attributes-v2.md
happy-barney: @mauke good point ... in fact, shouldn't `parse` be rather `Perl_keyword_plugin_t` ?

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+my $counter :shared;
+
+sub name :lvalue { ... }
+```
+
+The parser supports additional attributes that can be provided by modules, though the situation here has many shortcomings:
+
+* Only lexical or package variables, and subroutines, support attribute syntax. It is not possible to declare attributes on other entities such as packages, or subroutine parameters.
+
+* Attributes on anonymous subroutines are invoked only at "clonecv"-time; the time when an anonymous function gets turned into a closure with variable captures. It cannot perform any behaviour before this time.
```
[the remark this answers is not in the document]
leonerd: Ahyes - it seems I remembered it wrong. I knew it was only one one side, but apparently I forgot which. Will reword that.

ppcs/ppcTODO-attributes-v2.md
leonerd: Ahyes, good idea. It could just return the SV itself or it could return some new one. Would have to think a bit about refcounts, but yes it shouldn't modify the one passed in.

ppcs/ppcTODO-attributes-v2.md
leonerd: > Since most of these entities (lexical variables, anonymous closures) only exist at runtime, it would mean providing a runtime hook that is invoked when the entity is instantiated.
> 
> But that's a very different proposal.

Well, quite ;) That's the "Hooks" part of the idea, which is separate enough (and rather more vague) to warrant its own separate PPC doc. The attributes here aren't directly related, but they're likely useful when combined.

leonerd:
Many small updates clarifying wording, narrowing down a few open issues, adding more specification in terms of new API support functions and behaviours.

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+When the parser is parsing perl code and finds an attribute declaration attached to some entity, it can immediately inspect the lexical pad (and recurse up to parent scopes if applicable) in an attempt to find one of these lexical definitions. The first one that is found is invoked immediately, before the parser moves on in the source code. If such an attempt does not find a suitable handler, the declaration can be stored using the existing mechanism for a later attempt via the previous implementation.
+
+When attached to a package variable or package-named function, the target can be the entity itself (or maybe indirectly, its GV). When attached to a lexical, it is important to pass in the abstract concept of the target in general, rather than the current item in its scope pad. There is currently no suitable kind of SV to represent this - this remains another interesting open issue.
+
+Additionally, new callsites can be added to the parser to invoke attribute callbacks in new situations that previously were not permitted; such as package declarations or subroutine parameters.
```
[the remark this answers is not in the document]
leonerd: These are all things to consider in the future, yes. The point of this particular proposal isn't to allow new syntax, but to create a better implementation base such that in future new syntax could be considered.

I've added words to that effect.

ppcs/ppcTODO-attributes-v2.md
leonerd: I believe I've added this now in the latest wording. The pointer value is supplied as an argument to `newSVinternal_attribute_definition()`, and the callback functions in the struct all extended to use it.

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,236 @@
...
+An optional second callback function can be provided, for the purpose of parsing the incoming text from the source code into a value that the main apply function will use. Experience with the meta-programming layer in `Object::Pad` suggests this is useful, as often when meta-programming it is inconvenient to have to represent the parameters to an attribute as a flat text string.
+
+Pointers to these two functions are found by a structure that is associated with the name of the new attribute, perhaps defined as the following (though exact type is still an open issue; see below):
+
+```c
+struct PerlAttributeDefinition
+{
+    U32 ver;
+    U32 flags;
+    SV * (*parse)(pTHX_ const SV *text, void *data);   /* optional, may be NULL */
```
happy-barney, line 63: what about change `parse` to `custom_grammar` returning something that may or may not be `OP *` ? with following behaviour:

- when perl starts:
  - push grammar stack `unbalanced` grammar
  - push grammar stack `perl` grammar
- when encountering places where switch to known custom grammar is needed
  - push grammar stack `known custom grammar` (custom grammar can be `perl` as well)

there already is similar mechanism in `toke.c` - `Perl_lex_start`. This mechanism will just provide generic API to that plus it will provide `Perl` grammar itself as custom grammar.

That itself allows to implement `use VERSION` as custom grammar which may allow to introduce changes will require incompatible changes to `perly.y`.

Example: hypothetical import syntax may reuse this attribute mechanism for itself (by introducing custom attribute context)
```
use builtin
  :functions (
    foo :as (bar)
    baz
  )
  :attributes (magic)
  ;
```

ppcs/ppcTODO-attributes-v2.md
```diff
@@ -0,0 +1,236 @@
...
+    void (*apply)(pTHX_ SV *target, SV *attrvalue, void *data);
```
happy-barney, line 64: what about replacing this with some set (linked list? tree? ...) of structures:
```
struct attribute_apply {
    U32 attribute_context_id;
    Attribute_Apply_Function apply;
};
```

leonerd:
There's not been any new comments on this for a while, so I'm going to declare it ready-for-review. We'll see how it looks.

leonerd:
I've now added some more notes to outline a possible idea on what hooks might be, just to add some context for this doc. This isn't intended to be the place to discuss that idea in detail; save that for the other doc. It's just here to help evaluate how attributes might interact with them.

PR #57, opened 2024-11-15
Undef aware equality

leonerd:
Currently this draft PR contains two alternative suggestions, as raised by various discussions on perl5-porters@. We can continue to discuss the relative merits here as a draft, until we eventually decide to reject at least one of them, at which point the other will stand as a proposal to be submitted.

### DIFF
--- a/ppcs/ppcTODO-metaoperator-flags.md
+++ b/ppcs/ppcTODO-metaoperator-flags.md
@@ -1,0 +1,166 @@
+# Meta-operator Flags to Modify Behaviour of Operators
+
+## Preamble
+
+    Author:  Paul Evans <PEVANS>
+    Sponsor:
+    ID:      TODO
+    Status:  Exploratory
+
+## Abstract
+
+Defines a new syntax for adding behaviour-modifying flags to operators, in order to extend the available set of behaviours.
+
+## Motivation
+
+This idea came out of discussions around the proposal to add new `equ` and `===` operators, which are aware of the special nature of the `undef` value. The idea is that rather than adding just two new special-purpose operators, a far more flexible and extensible idea may be to allow syntax for putting flags on the existing operators.
+
+The central idea here is that existing operators can be modified by a set of single-letter suffix flags, that could be considered slightly similar to the concept of regexp pattern flags.
+
+## Rationale
+
+(explain why the (following) proposed solution will solve it)
+
+## Specification
+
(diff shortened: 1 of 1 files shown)


exodist:
I am 100% in favor of either one. I think the one that lets you use :u and :i is ideal. But I would be satisfied with the other one too.

guest20:
Feels like it should be a `/`, like regex modifiers ... but I guess that's taken by regex modifiers.

leonerd:
> Feels like it should be a `/`, like regex modifiers ... but I guess that's taken by regex modifiers.

There is an entire section in the document under "Rejected Ideas" explaining why it can't.

ppcs/ppcTODO-metaoperator-flags.md
```diff
@@ -0,0 +1,166 @@
...
+## Rationale
+
+(explain why the (following) proposed solution will solve it)
+
+## Specification
+
+Following the name of a regular value-returning infix operator, an optional set of flag letters can be supplied after a separating colon:
+
+```perl
+$x eq:u $y
```
rabbiveesh, line 29: I added prospective support for highlighting this to tree-sitter-perl; looks pretty solid, actually
image

Grinnz:
I think either option is nice, but I slightly favor the "modifiers" version:

If nothing else, it's a slightly different spelling of the other version that makes it clearer what's going on: eq:u vs equ is more obviously an eq operator.

If someday we decide the semantics are non-ideal we can fix them by just adding another modifier rather than a whole new operator, though this wouldn't be a huge problem other than spelling.

And least importantly, we can add other modifiers. not very important because there aren't any others that are obviously useful at this time. But maybe with the option available, the need will present itself.

Grinnz:
As a practical matter to what semantics we might change our mind on, I could go either way on whether these operators should consider undef equal to itself.

xiaoyafeng:
I  personally prefer modifiers one. Defined or not is a important concept in Perl,  But perl lack a neat and consistent way to deal with it. perl already has defined, undef, // and maybe ppc0021. but what all does just makes "defined or not"  more scattered in perl.   I think modifier is a good way to solve this problem.

guest20:
@leonerd do the debates still stay in github issues?

leonerd:
> @leonerd do the debates still stay in github issues?

That's a wider process question - perhaps @book @haarg @ap as the current PSC could comment on that.

PR #58, opened 2024-11-18
PPC: Perl version number

ap:
What it says on the tin.


ppcs/ppc0025-perl-version.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Security Implications
+
+There is a very marginal possibility that  code which does not expect a Perl revision other than 5 might present an opportunity for some kind of exploitation.
+
+## Examples
+
+`use 42;`
+
+Note that a v-string is not used. A v-string is how you can write `use v5.40` rather than the less readable `use 5.040`, where the important part is noisier – namely the version, 40, rather than the more prominent revision, 5. Without the revision, the version becomes the integer part and thus no longer requires padding. The subversion becomes the fractional part behind the first dot, with no need for a second dot – however, specifying a subversion on the `use VERSION` line no longer comes up anyway, because feature bundles are not subversion-specific and Perl point releases have long since lost the longevity and prominence they had in the 5.6/5.8/5.10 era. (If you were to need to specify a subversion, then without a v-string you would still have to write e.g. `use 42.001`, which is not as nice as `use v42.1` – but this is now academic.) Thus the need for a v-string goes away in practice. This also means Perl novices are no longer immediately confronted with them.
```
Grinnz, line 111: "It is significantly easier to remember that, hypothetically, use v7 will enable warnings and use v8 will enable signatures, than to remember that use v5.12 enables strict and use v5.16 enables __SUB__."

- https://blogs.perl.org/users/grinnz/2020/08/perl-7-by-default.html

That is, whether by 7 or 42, reclaiming the major version number gives us a nicer vehicle for representing new features. (i.e. I agree)

ppcs/ppc0025-perl-version.md
```diff
@@ -0,0 +1,165 @@
...
+
+1. The environment variable superseding leaves us with an awkward priority cascade for `PERL_LIB` > `PERL5LIB` > `PERLLIB`. However, the last one of these is already mainly historical.
+
+   Maybe we want to reuse `PERLLIB` by flipping the priorities instead of introducing `PERL_LIB`. In that case maybe the counterpart to `PERL5OPT` should be named `PERLOPT`, not `PERL_OPT`, given that these two variables are somewhat complementary. But this would miss the opportunity to tidy things up into “everything is named `PERL_*` (except `PERLIO`)”.
+
+   Maybe we want to just remove `PERLLIB`.
+
+   Maybe we just shrug and accept the awkwardness.
+
+1. Something needs to be done about version.pm:
```
Grinnz, line 149: Corollary: '42.1' parses as 42.100.0, because you are no longer using two dots. This might be a problem, and we cannot change how these numbers get parsed as versions.

ap: Yeah, that’s not changing. As acknowledged elsewhere, the next point release will (in `$]` parlance) be 42.001.

Grinnz: I was more stating that (as this section made clearer) we might run into additional confusion if we start trying to popularize 2-segment version numbers which will mostly break if we forget the v.

guest20:
@book `use warnings; use strict` is advice that still works 20 years after the first line of perl I wrote ... If I'd been given that advice back then I'd still be telling people to "use five eight eight"... 

I don't think a moving version number makes for a very good mantra, even if it's just a single digit...

ppcs/ppc0025-perl-version.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Preamble
+
+    Author:  Aristotle Pagaltzis <pagaltzis@gmx.de>
+    ID:      ARISTOTLE
+    Status:  Draft
+
+## Abstract
+
+There is no Perl&#160;6 looming any more, so the only thing “Perl” now refers to is Perl&#160;5. That has also forever been the language people mean when they say just “Perl” rather than specifying “Perl&#160;5”. There were earlier versions of Perl, but they are ancient history. All the history that created a distinction between “Perl” and “Perl&#160;5” has fallen away while nobody was paying conscious attention.
```
karenetheridge, line 11: grammar nit:

```suggestion
There is no Perl&#160;6 looming any more, so the only thing “Perl” now refers to is Perl&#160;5. That has also forever been the language that people intend when they say just “Perl” rather than specifying “Perl&#160;5”. There were earlier versions of Perl, but they are ancient history. All the history that created a distinction between “Perl” and “Perl&#160;5” has fallen away while nobody was paying conscious attention.
```

ppcs/ppc0025-perl-version.md
```diff
@@ -0,0 +1,165 @@
...
+
+In all this meantime it became clear that not everyone was going to adopt the new language, and that the ecosystem of the existing language was going to stick around. There would continue to be interest in Perl&#160;5 evolving for the foreseeable future, even as Perl&#160;6 came into its own. Thus Perl became a family. And that was now its future.
+
+Both conceptions of Perl’s future posed problems for the outside perception of Perl&#160;5. Even in the language family framework, though Perl&#160;5 was no longer the eventually-obsolete version, it appeared as the older, wartier, outmoded flavor. (In fact, both conceptions of Perl’s future equally posed problems for the outside perception of Perl&#160;6.)
+
+The community too, mirroring the versions-then-languages themselves, bifurcated, then separated, becoming two somewhat overlapping but distinct communities.
+
+And then, in October 2019, [Perl&#160;6 changed name](https://github.com/Raku/problem-solving/pull/89#pullrequestreview-300789072). Suddenly, it was no longer the future of Perl, in any shape, not even as the more advanced flavor.
```
karenetheridge, line 39:
```suggestion
And then, in October 2019, [Perl&#160;6 changed names](https://github.com/Raku/problem-solving/pull/89#pullrequestreview-300789072). Suddenly, it was no longer the future of Perl, in any shape, not even as the more advanced flavor.
```

book:
> @book `use warnings; use strict` is advice that still works 20 years after the first line of perl I wrote ... If I'd been given that advice back then I'd still be telling people to "use five eight eight"... 

Starting with `v5. 36`, in just one line, one gets a much better Perl than `use strict; use warnings;` can give them.

> I don't think a moving version number makes for a very good mantra, even if it's just a single digit... 

It's not about running after a moving target. It's about consciously picking which version of the Perl *language* a given program is written in.

ppcs/ppc0025-perl-version.md
happy-barney: it will make `use v42` ambiguous

haarg: Ambiguous how?

happy-barney: ok, not clear wording ... it will make it more often ambiguous for inexperienced reader (question: do we want to attract them?)

Yes, this one is already presented in perl:
```
package v42 {
}
```
but `use v42` fails

anyway, v7 approach, how it was presented by Sawyer, was imho very neat ... v7 variant will provide different set of features enabled/disabled/deprecated/obsoleted.

haarg: How is that any different between v7 and v42?

happy-barney: I know that, you know that, does newbie know that ?

What I'm trying to say: we should also weight learning curve and avoid possible ambiguities as much as possible. (yet maybe with over-complicated words)

book:
> when touching behaviour of `use VERSION` rather consider change from "use at least this version of interpreter" to "use exactly this version of syntax and behaviour"

You mean... break backwards compatibility?

Tux:
> > when touching behaviour of `use VERSION` rather consider change from "use at least this version of interpreter" to "use exactly this version of syntax and behaviour"
> 
> You mean... break backwards compatibility?

This is exactly what Jesse envisioned when he was Pumpkin. Interpretation and expectation also scattered then

I am in the camp of "Use this or newer", but I understand the other side

happy-barney:
> > when touching behaviour of `use VERSION` rather consider change from "use at least this version of interpreter" to "use exactly this version of syntax and behaviour"
> 
> You mean... break backwards compatibility?

please explain why do you mean it is breaking backwards compatibility?

I'm not saying that only `perl-5.12` will satisfy `use v5.12`, I'm saying that for example even with `perl-5.40` this will become syntax error:
```
use v5.12;
package Foo v1.0 {
}
```

simply because `package-version` was added in `v5.14`

happy-barney:
ok, this probably requires larger discussion, so I'll simply starts with some points I treat as pros:
- at long run simplifies core - maintaining grammar per version, callbacks per version, version id in XS symbols
  - allows to move "really old" versions into libraries
- allows incompatible changes (like `use fields` in 5.10)
- allows "forward" grammars - you should be able load newer grammar via library (and accepting related performance penalty)
  - so developer can start to use newer syntax even when production uses old binary with just newer library
- once implemented (and backported to older perls), there will be no need to address multi-version index

book:
> > > when touching behaviour of `use VERSION` rather consider change from "use at least this version of interpreter" to "use exactly this version of syntax and behaviour"
> > 
> > 
> > You mean... break backwards compatibility?
> 
> please explain why do you mean it is breaking backwards compatibility?

Sorry, I misread "use exactly this version of syntax and behaviour" as "accept only this version of the perl binary". Apologies.

> I'm not saying that only `perl-5.12` will satisfy `use v5.12`, I'm saying that for example even with `perl-5.40` this will become syntax error:
> 
> ```
> use v5.12;
> package Foo v1.0 {
> }
> ```
> 
> simply because `package-version` was added in `v5.14`

This is an error:

```
$ perl5.40.0 -e '
use v5.12;
package Foo v1.0 {
}
'
Invalid version format (dotted-decimal versions require at least three parts) at -e line 3, near "package Foo "
syntax error at -e line 3, near "package Foo v1.0"
Execution of -e aborted due to compilation errors.        
```

but this isn't:

```
$ perl5.40.0 -e '
use v5.12;
package Foo v1.0.0 {
}
'                                                                                         
```

ap:
@happy-barney 

> I'm not saying that only `perl-5.12` will satisfy `use v5.12`, I'm saying that for example even with `perl-5.40` this will become syntax error:
> 
> ```
> use v5.12;
> package Foo v1.0 {
> }
> ```
> 
> simply because `package-version` was added in `v5.14`

That’s not what `use v5.12` means now so it cannot be imposed retroactively. It is automatically going to be true in future for anything introduced behind a feature flag. Should everything new live behind a feature flag? That’s a debate I think. There are arguments in favor of that.

However, that is a separate topic from the version of Perl.

ppcs/ppc0025-perl-version.md
ap: Then I guess it’s good that newbies will be encouraged to just write `use 42` and not told about `use v42` at all.

It’s silly to write that anyway. There’s an extra “v” that changes nothing about the meaning except making it complicated to understand that it means the same thing.

The only reason we have v-strings is that we insisted for so many years on putting a redundant “5.” in front of what is really the version, and that made the version uglier to write as a number, so we went and invented a whole new kind of literal just to get rid of the noise.

happy-barney:
@book I have a feeling that you didn't understand what I'm trying to say :-( ... when you say "this code runs on 5.12 or more", it must run on 5.12. ... it should fail with every perl (that will also diminish need for perl version matrix testing necessity)

haarg:
> @book I have a feeling that you didn't understand what I'm trying to say :-( ... when you say "this code runs on 5.12 or more", it must run on 5.12. ... it should fail with every perl (that will also diminish need for perl version matrix testing necessity)

This is unrelated to this PPC.

duncand:
I fully agree that having "use 42;" is the best way to go for Perl version declarations going forward. No 5.x or 7.x etc, just X, and no "v" either. As to whether a declaration blocks use of Perl versions greater than 42 and some new syntax means 42-plus, that shouldn't be done unless something similar is supported for module versions, and that's all a separate and complicated discussion. The "use 42" should mean the same way of understanding as "use 5.040" etc does.

oodler577:
I vote no.

esabol:
I'm also not in favor of this PPC, at least in the specifics. I would be in favor of Perl's major version jumping to 7 in a near-future release, however.

happy-barney:
unrelated to previous comments in this PR, I'd mention my earlier proposal to switch versioning. Combined with this, what about:
```
use v2024;
```

also, pros and cons, what about treating these as aliases (ie, `v2025` will be an alias for `v5.42` ... both will be usable

book:
> unrelated to previous comments in this PR, I'd mention my earlier proposal to switch versioning. Combined with this, what about:
> 
> ```
> use v2024;
> ```
> 
> also, pros and cons, what about treating these as aliases (ie, `v2025` will be an alias for `v5.42` ... both will be usable

See also [Modern::Perl](https://metacpan.org/pod/Modern::Perl#Forward-Compatibility), which has been doing this [since 2012](https://metacpan.org/release/CHROMATIC/Modern-Perl-1.20120105/view/lib/Modern/Perl.pm#Forward-Compatibility).

book:
> this seems to coincide with some significance some people place on the number 42 and the upcoming opportunity the "5.42" version seems to represent;

`42` is exactly what you describe: a coïncidence.

book:
Quoting <https://semver.org/>, for those who are into it:

> # Semantic Versioning 2.0.0
>
> ## Summary
>
> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes
>
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

And noting that all PSC have proclaimed the same thing since their inception in 2020:

* in [2022](https://blogs.perl.org/users/psc/2022/05/what-happened-to-perl-7.html)
  > Existing sensibly-written Perl 5 code should continue to run under future releases of Perl. 
* in [2024](https://drive.google.com/file/d/1pQJfIW0u-4gKw1o-f18GyyPdT3YlwrUv/view) (check page 6, "Future Plans for Perl")
  > Don’t break things unnecessarily

I follows that:

1. We won't break Perl in the SemVer sense (or any other sense) in the foreseeable future.
2. Therefore, we don't need a MAJOR version number ([`PERL_REVISION`](https://github.com/Perl/perl5/blob/v5.41.6/patchlevel.h#L40) in the Perl source).
3. ~Profit!~ And so, we drop the 5.

ap:
> `42` is exactly what you describe: a coïncidence.

Although a lucky one, undoubtedly. 🙂

ppcs/ppc0025-perl-version.md
ap: I did some googling which confirmed my sense that using “mean” in this way is completely grammatical and standard, and I really don’t like the slightly different meaning of the version with “intend”. So I left out the “intend” bit of this suggestion.

I did however take the “that” bit – the sentence is indeed harder to parse without it. Good catch, thanks.

oalders:
I'm in favour of this change. +1

jberger:
I am also in favor, if it is the version number advance that will actually happen.

oodler577:
> > this seems to coincide with some significance some people place on the number 42 and the upcoming opportunity the "5.42" version seems to represent;
> 
> `42` is exactly what you describe: a coïncidence.

Sure, but someone noticed it and thought it'd be "cute" to generalize the "version" change, trying I am sure in charity to avoid the pitfalls of the "major version" bump to 7. "Just get rid of the major version" - problem solved. **Nope**.

You may recall that people also thought "Parrot" was cute when it was an April Fools joke, and we see how much pain that caused. Real frickin' funny 20 years later. We need to grow up and stop making decisions because they are cute or funny. 

Focus on the brass tacks. What problems would a better version scheme solve? What's the best version scheme to solve these problems? That's how you start, not fixating on a number like "42".

I noted elsewhere, that 6*7 = 42; so we're defining a series now. If this one fails, based on this formulata, then the version will be 1,764.

I prefer the existing semantic versioning. Just make the next release 7.0.0 with the same sort of conservative upgrade path that is done now, just call it 7.0.0 instead of 5.42.0. Don't require version bundling or "use vXXX" and more than you do now. It's not the number change that blew up the last attempt to make a 7.0.0, it was everyone treating it like it was any different or more special than another version bump. /. headline should be "Perl 7.0.0 released, fully backward compatible with 5.41.1 ..." and move on from there with the technical debates now without the "version number "distraction.

ap:
> That's how you start, not fixating on a number like "42".

We really don’t care what the exact number is (as @book already said above). You’re just asserting your assumption as fact. Personally I am not just amused and delighted that this comes up for version 42 but also 5% annoyed that it will seem intentional, and a bit on the nose at that, when it was not at all.

> We need to grow up and stop making decisions because they are cute or funny.

We didn’t. You’re not the only adult around. (And is it an adult who keeps asserting assumptions this way?) We went with this idea because we realized that aside from being cute, it also makes a lot of real sense in all sorts of ways. They are all laid out at length in the PPC itself, so I won’t repeat myself here. Your suggestion is covered in the Rejected Ideas section along with the reasons for rejecting it, which your response does not even touch on, let alone address.

oodler577:
`42 = 6 * 7`

duncand:
@book said:
> Quoting https://semver.org/, for those who are into it:
> > Given a version number MAJOR.MINOR.PATCH, increment the:
> > 
> > 1. MAJOR version when you make incompatible API changes
...

I want to emphasize that a valid interpretation of semantic versioning is that the given criteria is for when you MUST increment a position, and that it is NOT the only time you can do so.

In particular, it is completely valid to increment MAJOR for any reason, particularly marketing, even if it is entirely backwards compatible. The only thing you can't do is NOT increment MAJOR when you do have a backwards compatibility break.

For Perl, incrementing MAJOR every year just says there MIGHT be a backwards compatibility break, and without that we are guaranteeing backwards compatibility.

duncand:
I strongly vote YES for this proposal. Jumping to major version 42 is sending the best signal, that Perl has had MANY improvements over a LONG period of time, since it was first called 5. In contrast I strongly disagree with using version 7, per the stated reasons for rejecting that. Also staying with 5.x is bad. The only proposed alternative I would support is naming the versions after years, eg Perl 2025; but I think version 42 is MUCH better. See eg Jetbrains for precedent for the YEAR.MINOR versioning approach. See many things including Java and Firefox and Postgres and lots more for incrementing major for every non-bugfix release.

Grinnz:
> See many things including Java and Firefox and Postgres and lots more for incrementing major for every non-bugfix release.

I think the precedent of these in particular is a strong argument for using version numbers in this way, despite feeling that 7 is more appropriate personally.

Grinnz:
Something minor for Open Issues perhaps: perldelta naming. I try to avoid programmatically parsing perldelta names as the format is rather awkward currently, but we will want to do something consistent and probably also alphanumerically increasing, and it cannot be "perl420delta" because that will conflict with existing files once we get to Perl 56.1.

Grinnz:
Also, hopefully any naive heuristics out there will be updated not to interpret Perl 60 as Raku.

book:
> Something minor for Open Issues perhaps: perldelta naming. 

Very good point. I suppose we'll have to find a better naming scheme for those (I must admit I never liked the current system).

Grinnz:
None of the core perldocs use them currently but is there any reason they couldn't use underscore separators?

zmughal:
I worry that the specifics of this approach would spend more time on technical changes when the effort could be allocated elsewhere. This seems like it would require work for toolchain, devops, and distribution system package maintainers with little technical benefit for all the spillover (when compared to an important change such as `default_inc_excludes_dot`).

I have no problem with marketing, but part of marketing is focusing on message and copy.

An alternative might be to keep the semantics of "version" wherever it is used the same (tarball, XS, perlvar), but coin a new word that can be used to bring focus on the significant changes between each `5.x` version — that's the core message that I think we want to convey with a change like this. The other message is that there is a commitment to careful backward compatibility while making these changes: an evolutionary approach.

A suggestion for that wording is "release". This could look something like:
```perl
use release 42; # equivalent to: use 5.42;
# or shorter:
use rel 42;
```
and possibly with a syntax change:
```perl
use r42; # analogous to the shorter: use v5.42;
```
and then people can legitimately call that version "Perl release 42" or "Perl r42". This should not cause confusion as previous mentions in the wild would say "5.x" which can be disambiguated from `v5.5.x`.

The non-syntax change has the advantage is that it could be distributed as a dual-use module (another point towards the backward compatibility message).

duncand:
@zmughal said:
> I worry that the specifics of this approach would spend more time on technical changes when the effort could be allocated elsewhere.

I disagree. The current proposal for v42 is very much worth the effort. It has also been a long time coming, the same idea has been proposed seriously for years. I feel your proposed alternative is significantly worse for Perl.

iabyn:
I *LOATHE* this proposal.

leonerd:
It seems this PPC was merged way faster than the usual pace of development of PPCs, well before I had got around to properly writing down my thoughts about it. Ah well; I shall at least try to write something up here in case it may be of historic interest later on.

Here goes:

Eh.. I'm not going to stop anyone, but also I'm not enthusiastic about it either. I won't spend any of my time looking at this, when I could do things I would judge more useful elsewhere.

I *suspect* someone will come up on some fairly fundamental technical reason why something doesn't work, and requires a lot of fixing of the code.
That might stop this then and there.

It might be useful marketing optics, it might not. I'm not going to push for it myself because I don't see it having huge benefit, but if others want to and it's not taking away anyone else's time that could be more usefully spent building actually useful features, I say; go nuts.

I see it kinda like repainting the shop door. It doesn't matter what colour the door to the shop is, if nobody wants to buy any of your products. Maybe a repainted door will look shiny and fresh and tempt new people in? ... Perhaps. Hard to predict until we do it. But if we don't have any nice products for people to buy, those people will wander in, take a look around, and leave again. So we'd best be prepared to show them some good offerings to make their time worthwhile.

I will be spending my time designing and building better products to give to people. But if someone else wants to pick up a paintbrush, go right ahead.

haarg:
> I _LOATHE_ this proposal.

Can you elaborate at all?

> It seems this PPC was merged way faster than the usual pace of development of PPCs, well before I had got around to properly writing down my thoughts about it. Ah well; I shall at least try to write something up here in case it may be of historic interest later on.

The status is still 'draft'. It's been merged but not accepted. Feedback on this PR is still very welcome.

leonerd:
> The status is still 'draft'. It's been merged but not accepted. Feedback on this PR is still very welcome.

Ah; fair. That would be me confusing the "merged PR" with "accepted underlying idea" status, that I try to remind other people about. Hoist, petards,...

In that case, I hope my comments will help sway opinion.

neilb:
I'm not a fan. What is the _benefit_, and _for whom_? Perl is not on the upswing, it's not going to suddenly start gaining a big market share, and _nothing_ we do is going to change that. Lots of the recent developments are making a better Perl for the people already programming in perl, and that feels like the right focus. In particular, the people programming in perl who aren't active on p5p.

After 30 years of 5.x, suddenly jumping to 42, and specifically to 42, feels a bit shark-jumpy. Bumping the version to 7 when we have stable new OO, signatures, and a bunch of the other goodies in PPCs, seems appropriate. Then people can just write `use v7;` to get that new baseline, and most of the time won't worry about 7.2, 7.4, 7.6, etc.

But then don't bump to 8 until there are a bunch more notable things. Having a sequence 42, 44, 46, etc makes it hard to remember which is the right one to go for.

leonerd:
:+1: to everything @neilb just said

MartijnLievaart:
Op 22-11-2024 om 16:18 schreef Matthew Horsfall (alh):
>
> I'd like to plus 1 most of what @leonerd <https://github.com/leonerd> 
> and @neilb <https://github.com/neilb> said.
>
> And for me specifically, I worry about adding new ENV variables that 
> behave like and precede the old ones if present having the end result 
> of there being more ways to configure things which just adds to 
> confusion/maintenance costs and doesn't really (in my mind) add a benefit.
>

Would that really be a large issue?


> And I may be in the further minority here, but I don't like the idea 
> of Perl 7 at all. We keep coming back to this "Let's change the 
> version, it'll show the world..." but I don't think that will ever pan 
> out, and costs us all more time and energy to discuss internally, 
> develop, market, maintain, and explain. We've already lost so much 
> time and patience to this argument, and here we are doing it once 
> again. If we were to renumber, I think as Neil says, do it when we 
> have something to offer (feature class fully done, stable signatures, 
> SSL in core, ...)
>

Fully agree.


> I'd rather we focus on improving Perl, not repainting it. I don't 
> think folks are avoiding Perl because of the number "5".
>

I think they do. Perl is seen at my place of $ORK as a dead language 
from long ago. Having a rebrand will send the message that Perl is 
alive, well and evolving.


HTH,

M4

guest20:
Sure seem to be lots of folks on here with the opinion "nobody cares about version numbers but please I'm begging you to not change the version number" - Weird. 

> I'd rather we focus on improving Perl, not repainting it. I don't think folks are avoiding Perl because of the number "5".

It's a huge shame that the only thing p5p has done on perl is this one version thing and nothing else at all /s

oodler577:
> I think they do. Perl is seen at my place of $ORK as a dead language from long ago. Having a rebrand will send the message that Perl is alive, well and evolving.

Show them https://metacpan.org/recent and all the conference YT content from the last year. More to come in Decemeber, there a lot of activity!

happy-barney:
> > I think they do. Perl is seen at my place of $ORK as a dead language from long ago. Having a rebrand will send the message that Perl is alive, well and evolving.
> 
> Show them https://metacpan.org/recent and all the conference YT content from the last year. More to come in Decemeber, there a lot of activity!

They once read that Perl is bad and no one will change that.

Related experience: I had a discussion once with a manager who told me that he will rather hire programmer who will write 1000 lines of Java if/else code than programmer who will write regular expression instead.

rjbs:
> I'd rather we focus on improving Perl, not repainting it. I don't think folks are avoiding Perl because of the number "5".

I have mostly stayed out of this, but I feel I should chime in, for better or worse.  I agree with @wolfsage here.  If we're to renumber things, I think that "use the PERL_VERSION more prominently than PERL_REVISION" is a better strategy.  But a strategy is meant to gain us something, and I don't think this does.  The "motivation" section begins:

> Perl is in need of a future.

I don't think a version numbering change grants it any more of one than it has now.  Cranking up the version number, which will add some cost to downstream code that thinks about perl versions, *might* get people to look at what's new in the language since they last thought about it.  My sincere question is:  If someone hasn't thought about Perl since Perl 5.10, will the changes they see since then make them want to start using it again?  Put aside the question about version number:  if everyone who liked writing in Perl in 2010, but hasn't since 2010, were to suddenly watch a very good presentation about "what's new in Perl since 2010", would they stop writing in Go or Python or Deno and start writing Perl again, or contributing to the core?

If not, what is the practical way in which we should expect the change in version number to change Perl's future?

I also acknowledge that it's unfair to say "put this work aside because implementing core iterators would be better".  It's different work, of a different scale, to be done by different people, and so on.  "Don't do X, do Y" is a frequent and unfair demand of people in open source projects who don't think about what is and isn't fungible.

Finally:  Look, I'm not spending hours of time each week on Perl development or project management anymore.  If the people who are very active think this is a great idea, I don't need to cry out for it to not happen.  I think the cost, while nonzero, is not enormous.  I worry that the people who will benefit are either imaginary (people who will see the blog post, read the delta, and switch to working in perl) or the ones doing the imagining.

oodler577:
> They once read that Perl is bad and no one will change that.

bad is not dead, @MartijnLievaart said "dead"

> Related experience: I had a discussion once with a manager who told me that he will rather hire programmer who will write 1000 lines of Java if/else code than programmer who will write regular expression instead.

Perl is a tool best in the hands of specialists, all roads lead back to focusing on improving the language for the regular users and practicioners - it's future is to empower individual software master craftsman, that's where the world is headed and the next niche opportunity for Perl

happy-barney:
> > They once read that Perl is bad and no one will change that.
> 
> bad is not dead, @MartijnLievaart said "dead"
> 
> > Related experience: I had a discussion once with a manager who told me that he will rather hire programmer who will write 1000 lines of Java if/else code than programmer who will write regular expression instead.
> 
> Perl is a tool best in the hands of specialists, all roads lead back to focusing on improving the language for the regular users and practicioners - its future is to empower individual software master craftsman, that's where the world is headed and the next niche opportunity for Perl

Oh, starting future visions? :-) Funny, I've already shared my vision of future in `PPI` today ... imho it is worth to start Discussion somewhere - share visions, clarify why; share experiences, and visions how missing language/tooling features will make live easier. And somewhere there, picking common cases, is future (and of Perl probably as well)

duncand:
I don't actually consider changing the Perl version numbering scheme to be essential.

But IF we are going to do it, then technical concerns would be the same whether we go with 7 or 42, they both break the assumption of 5.x.y.

And IF we are going to do it, then I feel going with the 42+ is much better as it both signifies Perl has evolved a lot in 30 years, and it removes any thought that any particular release is uniquely deserving of a bump to eg 7 because of some feature or other it has.

duncand:
I feel there are at least 2 very distinct decisions to make.

One is whether or not to have Perl MAJOR version numbers that don't begin with 5 at all, never mind what new versioning scheme is used for MAJOR. This moving from 5 is where the main technical challenges would be.

The other is, IF we stop using 5, THEN what would its replacement scheme be.

I would argue that the single most important aspect of a replacement scheme is that it is a monotonic increase of the MAJOR, which simply declares there is a new version and makes no value judgement on worthiness based on features etc.

There is never going to be a Perl version that is "worthy" of a new MAJOR without all subsequent versions also being equally worthy of it.

Plato:
I too am not crazy about this proposal. Changing the version is going to get attention whether we want it or not. 

> 3. Move to Perl 7 based on some headline feature [...] opening up a future debate about whether changes are big enough to warrant a bump, and creating indirect pressure to introduce big headlining features every so often in order to justify a bump to avoid a renewed perception of stagnation. [...] there will be strong temptation to including breaking changes within its scope, every single time a new one happens

But are there any downsides? Seriously, this is my favoured approach, and some of those pressures are going to happen anyway, and are not necessarily all negative.

duncand:
A few further thoughts, IF we go from 5.x ...

I agree with continuing the odd/even dev/release pairings, but they are MAJOR.

I propose taking both numbers 42 and 7 off the table for stable releases.

If we like the idea of 7, go with 8 instead, such that 7.x is the dev series and 8.x is the release series. Not only does this give us the odd/even pairing, but the number 7 now also has some baggage.

If we like the idea of 42 for respecting long history of development but not how 42 seems gimmicky due to its other meanings, go to 44 instead for release (and 43 for dev), then we avoid that gimmicky baggage. And I say skip meaning we can do it now rather than having to delay a whole year until it falls out naturally. Just skipping will hardly look different as they're all bigger numbers at that point.

ap:
@nailb

> What is the _benefit_, and _for whom_?

By far the primary benefit is settling the question of “when will Perl have another version”, with the perennial tag-along topic of “can’t we break compat while we’re in there”. The beneficiary is the Perl community.

A big secondary benefit is the fact that `use VERSION`, which we mean to put much more emphasis on in the future, becomes much easier to explain to newbies if it’s just `use 42` rather than `use v5.42`. The beneficiary is whoever learns Perl now, i.e. ultimately the Perl community. Part of this is that we want `use VERSION` to have more emphasis in the future generally, because we see that as a the way to move the language forward without a *need* for breakage. We can turn retroactively invented features *off* if you ask for a recent Perl version, so you can have a cleaner language without everyone else also having to rewrite their code to accommodate you in that.

A tertiary benefit is making everyone else aware that we weren’t sitting on our hands while the whole Perl&#160;6 thing was going on. The beneficiary is again the Perl community. At least whoever doesn’t treat this as a pure PR ploy; obviously there are going to be plenty of people who react to the news that way. But my own expectation is that in 5 or 6 years it will seem *bizarre* that we stuck with 5.x for so interminably long.

@wolfsage

> And I may be in the further minority here, but I don't like the idea of Perl 7 at all.

Neither do I. That’s *why* I think we should implement this PPC: to settle the question. (See my reply to Neil above.) As long as Perl remains Perl&#160;5, *someone* is always going to agitate about that. And not entirely for no reason either: people who have represented Perl at non-Perl-community tech conferences consistently report some fraction of outsiders think the next version of Perl just never happened so it’s just this old dusty language now.

@rjbs

> > Perl is in need of a future.
> 
> I don't think a version numbering change grants it any more of one than it has now.

I agree, it doesn’t. Instead it says “we already have the future we need”. What I mean wjen I say that Perl is in need of a future is our idea of the future of Perl. Like @wolfsage, I think we do not need another Perl&#160;6 nor another Perl&#160;7 nor whatever “let’s make a clean break with some stuff” might come next. But the “Perl&#160;5” framing suggests thinking of it in precisely that way. I think as long as we don’t bump the version *in some way*, the matter is going to keep coming up in some form or other. Perl needs a future, but that doesn’t mean we don’t need to create one, we just need to recognize that we already have it.

And bumping the version in *this* way is the absolute least nonzero technical effort way of bumping it, due to the fact that XS is completely unaffected by it.

The only lesser-effort way is not to change `$]`/`$^V` either, and just *call* it Perl 42 in marketing. But if the PR benefit of this PPC is questionable, the PR non-benefit of that approach is even worse.

And there is good reason to make an actual technical change, in that it cleans up `use VERSION`: no more v-strings. Just `use 42` instead of `use v5.42`; the former you can explain to a newbie, the latter they have to take on faith (and it’s ugly when they get around to digging into it). And we want to put emphasis on putting a `use VERSION` on your code, basically the way it was with use strict and warnings.

@duncand

> But IF we are going to do it, then technical concerns would be the same whether we go with 7 or 42, they both break the assumption of 5.x.y.

You might think that, but again, it is in fact *easier* to bump to 42 than to 7 (and by a lot) if done in this particular way.

duncand:
@ap 
> You might think that, but again, it is in fact _easier_ to bump to 42 than to 7 (and by a lot) if done in this particular way.

Yes, you're right, sorry. When I wrote that I was thinking specifically about the tool chain or repackager stuff that specifically assumed 5.x.

But what do you think about going to 44 rather than 42? It should be the same from a technical effort, and avoid gimmicks.

Grinnz:
> If we like the idea of 42 for respecting long history of development but not how 42 seems gimmicky due to its other meanings, go to 44 instead for release (and 43 for dev), then we avoid that gimmicky baggage.

I have not seen anyone express this as a downside that 44 would do better. But it is a good point that if this is agreed upon, we should start with a Perl 41 dev release.

ap:
@duncand

There are a few Perl community insider gimmicks that I would be sad to miss out on. The HHGTTG gimmick I like as well, actually; the fact that some people will say that’s why we did it is marginally annoying but not enough to sour my enjoyment of the other benefits. That’s just how I feel about the number though. The approach is what matters to me, if people agree that we should take this approach but go to something other than 42, I’m fine with that.

@Grinnz

A 41 dev release is mandatory if we’re going to do this. One of the reasons this PPC went in so quickly (👋 @leonerd) is so we have at least some time left for that.

(@leonerd It did spur us to look at the other PPC PRs and pretty much decide to merge most of them. Part of the problem is that we have too many epicycles in the PPC process; PPCs have statuses but then they are also proposed in GitHub PRs where there can be discussion and then there are also pre-PPCs on the list. It’s all a bit undefined as to which part of all this process serves what purpose, and parts of it are probably redundant. We spent some of the last PSC call on that subject.)

book:
> But it is a good point that if this is agreed upon, we should start with a Perl 41 dev release.

Most likely, 41.7 or 41.8.

Hopefully, by the time we actually do this, we'll be past cold feets, and will have thought of all the most problematic issues, so we don't have to revert.

book:
@neilb 

> Having a sequence 42, 44, 46, etc makes it hard to remember which is the right one to go for.

Isn't it already hard to remember which of 5.36, 5.38, 5.40 is the right one to go for?

jberger:
I hate the idea of discussing this to death like it has been so many times. 

I also hate the idea of "we should wait until there's something worth marketing as new". Folks, we have added so many things since 5.0 that are just begging for recognition as new. Nothing ever seems to be worth it to the "wait for something big" folks. So why not promote each new release? Isn't each release effort worth highlighting? (If not then lets set that benchmark someplace concrete so that we could eventually get to it someday).

Will a blog post about changing the version scheme get more people to try perl, no probably not? But will someone who doesn't know and looks to see "when was the last time Perl had a major release" and they see this year and not 1987 think better of it? I think they will; and before you ask, yes that comes up outside the echo chamber a LOT.

duncand:
> @neilb
>> Having a sequence 42, 44, 46, etc makes it hard to remember which is the
>> right one to go for.
>
> Isn't it already hard to remember which of 5.36, 5.38, 5.40 is the right
> one to go for?

I don't see what problem you are pointing to here.  What "right one to go
for"?

If it is a matter of users deciding which version of Perl to use, the
answer generally should be, whatever is the newest stable one at the time,
or else the closest to that which is justified staying back on.

Look at Firefox or Chrome, whose major versions increase every 6 weeks or
so, and are now like around number 132 or whatever. Are those high numbers
confusing people on what is the "right one to go for"?

If one is using a package manager in their operating system to install
Perl, it would probably be regularly kept up to date on the latest one
available without much thought.

oodler577:
> Are those high numbers confusing people on what is the "right one to go for"?

They confuse me because context (or _semantics_) is lost. I may just be fully unable to think outside of the semantic versioning (dotted number)  box, but any software that uses just major numbers always confuses me. Angular is a good example of a software stack that uses is really confusing with its versioning. For Firefox, it seems okay. Maybe for software backends it's not helpful. For user facing consumable software products it's probably good. Perl is the former.

happy-barney:
> I don't see what problem you are pointing to here. What "right one to go for"? If it is a matter of users deciding which version of Perl to use, the answer generally should be, whatever is the newest stable one at the time, or else the closest to that which is justified staying back on. Look at Firefox or Chrome, whose major versions increase every 6 weeks or so, and are now like around number 132 or whatever. Are those high numbers confusing people on what is the "right one to go for"? If one is using a package manager in their operating system to install Perl, it would probably be regularly kept up to date on the latest one available without much thought.

Unlike Firefox and Chrome, Perl (intended usage) is something that runs "business critical" code (meant - something that brings value). Newer version usually means extensive testing and code review - no one will play such frequent releases without guaranteed backward compatibility (and that Perl does not even now).

duncand:
> Unlike Firefox and Chrome, Perl (intended usage) is something that runs "business critical" code (meant - something that brings value). Newer version usually means extensive testing and code review - no one will play such frequent releases without guaranteed backward compatibility (and that Perl does not even now).

What you say is the same no matter what versioning scheme Perl uses. This whole "right one to go for" has nothing to do with the versioning scheme. That didn't seem to be pointing to an actual problem with the versioning scheme.

oodler577:
> That didn't seem to be pointing to an actual problem with the versioning scheme.

He is pointing out that there is no value added to the new scheme for people who make money and do important ("business critical") things with Perl. In fact, this change would be a net cost - much of it hidden and take out of Perl's own hide.

book:
> Newer version usually means extensive testing and code review - no one will play such frequent releases without guaranteed backward compatibility

Isn't Perl guaranteeing just that?

The magic of [feature](https://perldoc.perl.org/feature#FEATURE-BUNDLES) (and since v5.40, [builtin](https://perldoc.perl.org/builtin#Version-Bundles)) bundles is what really powers `use VERSION`.

When a Perl program starts with `use v5.36;`, the version number is less about the version of the `perl` *binary* (obviously, it has to be greater than or equal to 5.36.0) and more about the version of the Perl *language*.

The Perl Porters **promise** that `perl5.40.0` (the binary program) will understand exactly what that `use v5.36;` line means and do exactly what was expected by the code author. They actually promise that all future releases of `perl` will understand that, and behave accordingly.

This makes transitioning to a new version of a the *binary* much easier. At work, we jumped from using `perl5.28.1` to `perl5.36.0` without much stress. Most of our users definitely didn't notice the upgrade. We plan to move to `perl5.38.2` or `perl5.40.0` just as easily, without making a big fuss for our developers. We'll likely announce it *after* it's done.

Someone answer me this: *How often does one of the fundamental components of your company's software platform change without a big scary announcement, and a frightening deadline for the update?* (Which probably happens on a Sunday, to minimize disruption.) When we bump the version of our `perl` binary to some bigger one, it's going to be just another Thursday.

Some of the company code was written when `perl5.8.5` was current. It transitioned, without needing to be updated (or even looked at), through some flavours of v5.14, v5.24, v5.28, up to v5.36. Some of it still runs today, *unmodified*. Across all of our tens of thousands of Perl modules, over 30 different `use VERSION` lines are used. From v5.1 (someone typoed `5.010` as `5.001`) up to v5.36, going through v5.10, v5.18.2 (someone didn't know the sub-version is ignored), v5.24, etc. The word for this is "legacy".

The only real difficulty arises when changing the number on the `use VERSION` line in the code. And this is where the real (lexical) magic happens: modules and programs can be updated *one at a time*.

And that transition can be difficult:
* Perl v5.28 subtly changed the meaning of `|`, `&`, `^` and `~`
* Perl v5.36 won't understand `$fido = new Camel "Amelia";`
* Perl v5.36 will complain about `sub foo ($$);`
* Perl v5.38 will complain about `open FH, $file;`

But, as long as the `use VERSION` line at the top of the file matches the version of the Perl *language* that the author meant, `perl5.40.0` will run the code *exactly* as expected.

To be fair, it took us Porters some time to get to this point: the experimental `autoderef` was not guarded by a `feature` and its removal caused `perl5.24.0` and later to choke on any code using it, no matter which `use VERSION` line was used. I believe nowadays we (the Porters) better understand the tradeoffs of putting something behind a feature guard or not.

In all the above examples, the "5" in the `use VERSION` line brings exactly *zero* information. It's actually cluttering the line. By "dropping the 5" (actually the `v5.`), we make the other number (28, 36, 38), the only *meaningful one*. Picking this number is the only choice Perl programmers have to make when writing new code.

Picking exactly *which* version of the Perl *language* to use is the actual question one has to answer. And the answer will be different if one is a CPAN author, a hobbyist writing their own tools, an Open Source application developer or a developer for a company's web application or internal tooling. The answer depends on several factors: one of them is the programmer's desire to use recent Perl language features, and another is their expectations regarding the minimum version of the `perl` *binary* the code is expected to run on.

@rjbs wrote:
> If we're to renumber things, I think that "use the PERL_VERSION more prominently than PERL_REVISION" is a better strategy.

@ap wrote:
> And we want to put emphasis on putting a use VERSION on your code, basically the way it was with use strict and warnings.

By dropping the 5, we *are* making the `PERL_VERSION` number more prominent.

The apparent big version jump puts us in a position where not only we get to tell the world (at least those willing to listen) that Perl has never stagnated, and we can also articulate the importance of picking a *language version* to current Perl users (many of them need to finally understand what `use VERSION` actually means).

briandfoy:
> But, as long as the use VERSION line at the top of the file matches the version of the Perl language that the author meant, perl5.40.0 will run the code exactly as expected.

@book: This simply is not true. The `use v5.x` line does not make code from old perls work on new perls. The presence of `use v5.14` in this code does not mean the code works like it did under v5.14: 

```
$ perl5.24.0 -c -e 'use v5.14; my $ref;  push $ref, q(foo)'
Experimental push on scalar is now forbidden at -e line 1, at EOF
-e had compilation errors.
```

Many features are enabled by flags and the code to support them exists but may not one enabled. However, some things have been completely removed or fatalized. There are plenty of more examples.

Plato:
If this were to proceed, what would the press release say?

ap:
@guest20 Good catch, thanks. (There doesn’t seem much to figure out there, other than to just do it, so it goes straight into the Specification instead, but same difference.)

book:
> > But, as long as the use VERSION line at the top of the file matches the version of the Perl language that the author meant, perl5.40.0 will run the code exactly as expected.
> 
> @book: This simply is not true.

@briandfoy: You are, of course, absolutely right: I was lying when I said `feature` was magical.  :sweat_smile:

I noted it myself, when I mentioned the `autoderef` feature. Not every new Perl feature is guarded by a `feature` (see [`Syntax::Construct`](https://metacpan.org/pod/Syntax::Construct) for an extensive list).

In both our examples, though, the troublesome construction was *experimental* when it was introduced, and it was removed without having ever left the experimental state. I'm not convinced that code that uses experimental constructs falls under the definition of "sensibly-written" (as written by the 2021-2022 PSC in [What happened to Perl 7?](https://blogs.perl.org/users/psc/2022/05/what-happened-to-perl-7.html)).

Nowadays, p5p would not introduce a change of meaning or behaviour in valid syntax without a feature guard (for an example, see the [`bitwise`](https://perldoc.perl.org/feature#The-'bitwise'-feature) feature, which changed the meaning of the binary operators). See also how the removal of deprecated features is handled, most recently with the [`apostrophe_as_package_separator`](https://perldoc.perl.org/blead/feature#The-'apostrophe_as_package_separator'-feature) feature.

We're aiming for a future where most of the old ("sensibly-written") code will continue to work, untouched, with the latest `perl`, while everyone can write better Perl by pointedly picking the best `use VERSION;` line for their use case.

My message was driven by my enthusiasm for `use VERSION` itself. No matter if this PPC goes through or not, we want to make `use VERSION` the staple of modern Perl code. Yes, Perl is incredibly backwards-compatible. As we've seen, though, that's not always 100% foolproof. `use VERSION` is still the best hint a programmer can give to their language interpreter/compiler to ensure everything will work as expected, for as long as possible.

One of the points of this PPC (besides "Perl 5 is Perl" and "there is no successor to Perl"), is that `use 42;` is much more readable (and less error-prone) than `use v5.42;` or `use 5.042;`.

iabyn:
On Fri, Nov 22, 2024 at 04:54:10AM -0800, Graham Knop wrote:
> > I _LOATHE_ this proposal.
> 
> Can you elaborate at all?

There are two sides to this issue: technical and marketing.

From the marketing side, I think that changing the versioning scheme will
have exactly 0.0000% impact on slowing the gradual decline of perl.
Most people outside the perl sphere wont notice or care, and the few who
do notice will (correctly) assume that this is just a cynical marketing
ploy.

From the technical side, there are no benefits. There are definite
downsides however. Perl versioning is very visible to the language,
e.g.

    use v5.40;

    if ($] > 5.020_001) { ... }

    #if PERL_REVISION == 5 && PERL_VERSION >= 6

etc, and the version can be expressed in lots of different ways, as shown
in the examples above. Changes to the scheme just increases the cognitive
load. We've only recently reached the point where we no longer have to
worry much about version 5.005_03 being a thing, and the transition to
5.6.0 was 25 years ago.

Consider: do people know off the top of their heads what each of the
following do:

    use v5.40;
    use v5.40.0;
    use v5.040000;
    use v5.040_000;

    use 5.40;
    use 5.40.0;
    use 5.040000;
    use 5.040_000;

I certainly don't.  Why add further confusion for apparently zero gain?

I remember when Sun decided that SunOS 5.x would be branded as Solaris
2.x; then they decided that Solaris 2.y would be rebranded as 'Solaris y',
but under the hood, uname etc would tell you it was 'SunOS 5.y'.

-- 
A power surge on the Bridge is rapidly and correctly diagnosed as a faulty
capacitor by the highly-trained and competent engineering staff.
    -- Things That Never Happen in "Star Trek" #9

guest20:
Paraphrasing @iabyn 
> v-strings, version.pm, feature bundles and `_` in numbers, oh my! 

It's definitely worth brushing up on those things† in a discussion about those things. I know I had to. 

> Why add further confusion for apparently zero gain?

With a two part version there is no need for the super confusing v-string binary-packed decimal notation in the first place, and since 42.1 fits into a regular old number we no longer need version.pm to make things comparable via overload jank.

With two part versions most of the things that make you uneasy about those version declarations magically go away‡, and that sounds like a pretty major gain to me. 

Now we can compare numbers like the Arabic scholars of old intended.

>    if ($] > 5.020_001) { ... }
>    #if PERL_REVISION == 5 && PERL_VERSION >= 6

This is covered in the PPC under "Backwards Compatibility". 

I can't see the rest of this hypothetical code, but I'm betting it's an early return using some feature from newer perl in the positive case, and some ancient jank in the fall-through case. 42 being greater than 5.020_001 seems to preserve those semantics pretty well. 

I will have to admit a weakness at this point and admit that although it does sound unnerving that `int $]` and PERL_REVISION might not be equal but I really don't do enough XS to be upset by that. I don't even know how to access $] in XS⸸ 

__ 
†.
https://xdg.me/version-numbers-should-be-boring/
https://perldoc.perl.org/feature#FEATURE-BUNDLES
https://perldoc.perl.org/perldata#Version-Strings
‡. ... except _, which really feels like it's being dragged into this against its will 
⸸. ... is it macros? I bet it's macros

leonerd:
On the subject of the v-string numbers specifically: I might be in the minority, but I actually *like* the leading v.

If you look at the start of a file and you see

```perl
use v5.36;
```

you probably have some idea that Perl version 5.36 is a thing. And that leading `v` gives more of a hint here as to what's going on. "version 5.36", yes that sounds like a related thing here.

```perl
use 42;
```

What's this number doing here? I get no clue as to what's going on.

Whatever actual number(s) we decide on in future, at the very least I suggest we keep that leading `v` in all the examples and suggested "here's how you should write your code", because it is actually useful.

ap:
@iabyn

Thanks for elaborating.

> Changes to the scheme just increases the cognitive load.

The thing that confuses me about your reaction is that we’re not proposing no new or altered cases in this scheme at all. We’re proposing changing only the *value* of the version, not any of the semantics of version numbers.

And *for users*, the proposal ends up removing all of that cognitive load, because it means the only case relevant to day-to-day business is `use INTEGER`. And if users ever do need to wade into this marsh, they will not have to deal with anything more than is already the case.

We do increase *our own* cognitive load in the one detail that `PERL_REVISION` is no longer part of `$]`/`$^V` from a certain `PERL_VERSION` on. This is relevant essentially only for the porters/toolchain crowd and XS authors, and only in such XS code that (thinks it) needs to cross-reference those two sources of truth for some reason. The reason to take on this (IMO minor) increase in cognitive load is that we get to simply decline to pay the transition cost of bumping `PERL_REVISION` – which is prohibitive and would otherwise prevent us from creating the huge cognitive load benefit for users. So the gain seems *easily* worth the cost.

Therefore, on the grounds you outline, it seems like I should expect you to love rather than loathe this proposal. If that still isn’t the case, where am I mistaken?

ap:
@leonerd

> What's this number doing here? I get no clue as to what's going on.

But… you personally do know, no? This preference just seems to me like optimizing for people who have never seen Perl code before – particularly in the context of the future we’re hoping for, where `use INTEGER` is the first thing every Perl novice learns and the first thing every Perl file starts with.

Plato:
> I'm betting a very similar thing to [any other release notes](https://www.nntp.perl.org/group/perl.perl5.porters/2024/09/msg268800.html)

I didn't mean the release notes, I meant the marketing copy. What's the official announcement going to say to anyone who happens to notice?

guest20:
@Plato Maybe you can point me at a previous example of some perl "marketing copy"? I can't personally think of anything that fits this description outside of the release notes...

oodler577:
Why after this entire discussion is this still in play?

HOW ABOUT WE TABLE TO DISCUSSION OF CHANGING VERSIONS FOR THE NEXT 100 YEARS. It never ends well, always results in extremely hurt feelings, and inevitably pushes more people. PLEASE PLEASE PLEASE STOP MESSING WITH THE VERSIONS!

https://www.nntp.perl.org/group/perl.perl5.porters/2024/12/msg269322.html

> 
> * We went over our findings so far regarding Perl 42. It looks like Perl
>   versions will have to continue to be written with three components, so
>   we will be stuck with a fixed .0 at the end. It's slightly annoying,
>   but we can live with it. We strategized about how to get more
>   real-world data about feasibility, and discussed the possibility of
>   releasing a perl-41.8 tarball alongside the regularly scheduled
>   perl-5.41.8.

**perl-41.8** !!!???

Do you have any idea how much confusion and more black eyes this will cause?

Grinnz:
Do you have any specific concerns to share?

happy-barney:
>
> With a two part version there is no need for the super confusing v-string
> binary-packed decimal notation in the first place, and since 42.1 fits into
> a regular old number we no longer need version.pm to make things
> comparable via overload jank.
>

What did you mean?
- v42.100 ? (as currently interpreted)
- v42.1 ? (as usual in versioning?)

Compare with other languages:
- which language as capability to say that in code ?
- which language supports work with versioning (eg: javascript's module
semver)

happy-barney:
On Mon, 25 Nov 2024 at 17:43, Aristotle Pagaltzis ***@***.***>
wrote:

> @leonerd <https://github.com/leonerd>
>
> What's this number doing here? I get no clue as to what's going on.
>
> But… you personally do know, no? This preference just seems to me like
> optimizing for people who have never seen Perl code before – particularly
> in the context of the future we’re hoping for, where use INTEGER is the
> first thing every Perl novice learns and the first thing every Perl file
> starts with.
>

Maybe not:
- number of novices looking into CPAN modules tends to zero
- proprietary codebases usually use some kind of "use
Project::Common::Sense" which enables all necessary features including
version

guest20:
@oodler577 neat English Fact™: To "table" an item means two different things depending on the dialect of the reader. 
- In British (and "most" Englishes) it means "bring up in discussion". 
- In American it means "remove from discussion".

guest20:
@happy-barney said, non-at-mentioningly: 
> What did you mean?
> - v42.100 ? (as currently interpreted)
> - v42.1 ? (as usual in versioning?)

No, neither of these, because these are both v-strings, and the version will now fit neatly into a float, like a number. `42.1`, that is fourty two and one tenth. Not the 2-character string of ascii 42 followed by ascii 1 (or acsii 100) as the v-strings you suggested here would produce. It would be a regular float, and one very far away from the precision limits for floats. 

You are asking "What about this confusion we have because we needed v-strings to represent 3 part versions? How does your two part version deal with this problem?". The answer is, "It addresses it directly by actually removing the problem". v-strings will only be useful for neat perl golfs and upsetting newbies on IRC. Possibly as a fun prank for binary packing ip addresses too?

Grinnz:
On Mon, Dec 23, 2024 at 5:44 AM guest20 ***@***.***> wrote:

> @happy-barney <https://github.com/happy-barney> said, non-at-mentioningly:
>
> What did you mean?
>
>    - v42.100 ? (as currently interpreted)
>    - v42.1 ? (as usual in versioning?)
>
> No, neither of these, because these are both v-strings, and the version
> will now fit neatly into a float, like a number. 42.1, that is fourty two
> and one tenth. Not the 2-character string of ascii 42 followed by ascii 1
> (or acsii 100) as the v-strings you suggested here would produce. It would
> be a regular float, and one very far away from the precision limits for
> floats.
>

How would this work for the version after 42.9? There is a reason versions
are not floats in any other software project.

-Dan

>

guest20:
@Grinnz The exact thing that's already in place when we put version numbers into floats. A leading zero: as seen in `use 5.020`. 

> There is a reason versions are not floats in any other software project.

Is it because people want major, minor and point? I bet it's because people want major, minor and point.

Grinnz:
> @Grinnz The exact thing that's already in place when we put version numbers into floats. A leading zero: as seen in `use 5.020`.

Then it would be 42.001, not 42.1... hence my question.

guest20:
@Grinnz I guess that was the actual question @happy-barney was asking in the first place too. Thanks for the nudge.  

I didn't even consider that this might have been the question because numbers just can't work like that.

So yes, absolutely, `42.1` isn't less than `42.100`, for floats to represent versions you gotta zero pad 'em

happy-barney:
@guest20 yes, that was my question ... and again I don't understand what I did wrong that you had to guess :-(

guest20:
@happy-barney Since @Grinnz managed to interpret your question correctly and it's very clear what you're asking now I re-read it I can only conclude that the problem wasn't what you wrote, but with what I read. Maybe I got thrown by your use of v-strings in english text, but either way, that's a skill issue on my part.

I guess my tone must have come off pretty shitty to get a `:-(` from someone with "happy" in their handle/avatar, so sorry for that too.

happy-barney:
@guest20 I do observe quite frequently that my text is not understand as I meant it (trying to find pattern in it and improve), that `:-(` was for me

As non-native speaker I do not recognize much of tones in English text (or maybe my brain is wired different ... still same result).

oodler577:
Given that the tone of this open thread was largely net-negative regarding adding confusion to Perl versioning (which this move does), it appears to be a break in the ranks that I hope will be addressed as an internal matter. My only recourse is to appeal to Perl influencers who are against this but may be withholding their opinions. Releasing a "perl-41.8" would be a significant mistake, and if I can't appeal to the right people to prevent it, then I hope the resulting fallout is severe enough that those responsible face irreparable damage to their reputations and quietly step away.

book:
> Why after this entire discussion is this still in play?

And why reply to a closed PR? I believe the discussion you want to have would be more visible if you opened a new issue on the repository, of you replied directly to the mailing-list post you're quoting.

> https://www.nntp.perl.org/group/perl.perl5.porters/2024/12/msg269322.html
> 
> **perl-41.8** !!!???

Anyone can push a tarball named perl-$version to CPAN (there was a small security scare a few month ago when someone did exactly that). We intend to make sure this won't be indexed as an official perl release.

There will be a regular perl-5.41.8 tarball, released as usual. The point of making a perl-41.8 is to experiment. We intend to go as far as possible with the experiment, to get the best possible understanding of what the technical issues with it are.

guest20:
Nah, replying to this thread it goes to anybody who's commented here, a new issue only goes to the owners of the repo on github

book:
So maybe don't cross the streams and reply to the list?

PR #60, opened 2024-11-25
Clarifications on PPC0027

leonerd:
Closes the existing "open issues" with regard to naming of the feature flag.

Also opens one new issue to do with aliasing vs. copy of `$_`.

### DIFF
--- a/ppcs/ppc0027-any-and-all.md
+++ b/ppcs/ppc0027-any-and-all.md
@@ -116,6 +116,8 @@ In any case, as junctions behave like values, they do not require special syntax
 
 ## Open Issues
 
+* A little-used behaviour of `grep` and `map` is that the `$_` variable does not merely store a copy of each original list element but actually aliases it. This is almost never used intentionally and can often lead to accidentally modifying the original list values, leading to subtle data corruption bugs. As these proposed operators are new, there will be no compatibility issues in changing the behaviour here to say that they are copies, not aliases. Perhaps they should even be read-only copies.
+
 ## Copyright
 
 Copyright (C) 2024, Paul Evans.
--- a/ppcs/ppc0027-any-and-all.md
+++ b/ppcs/ppc0027-any-and-all.md
@@ -19,9 +19,11 @@ Most code of any appreciable size tends to make use of at least the `any` or `al
 
 ## Specification
 
-New named features that, when enabled, activate syntax analogous to the existing `grep` operator, named `any` and `all`:
+A new named feature that, when enabled, activates new operators analogous to the existing `grep` operator, named `any` and `all`:
 
 ```perl
+use feature 'list_utils';
+
 any { BLOCK } LIST
 
 all { BLOCK } LIST
@@ -50,9 +52,9 @@ Some::Class->new(
 
 ## Backwards Compatibility
 
-As these new operators are guarded by named features, there are no immediate concerns with backward compatiblity in the short-term.
+As these new operators are guarded by a named feature, there are no immediate concerns with backward compatiblity in the short-term.
 
-In the longer term, if these named features become part of a versioned feature bundle that is enabled by a corresponding `use VERSION` declaration there may be concerns that the names collide with functions provided by `List::Util` or similar modules. As the intention of these operators is to provide the same behaviour, this is not considered a major problem. Differences due to caller scope as outlined above may be surprising to a small number of users.
+In the longer term, if this named feature becomes part of a versioned feature bundle that is enabled by a corresponding `use VERSION` declaration there may be concerns that the names collide with functions provided by `List::Util` or similar modules. As the intention of these operators is to provide the same behaviour, this is not considered a major problem. Differences due to caller scope as outlined above may be surprising to a small number of users.
 
 ## Security Implications
 
@@ -60,7 +62,7 @@ In the longer term, if these named features become part of a versioned feature b
 
 ```perl
 use v5.40;
-use feature 'any';
+use feature 'list_utils';
 
 if( any { $_ > 10 } 5, 10, 15, 20 ) { say "A number above 10" }
 ```
@@ -113,10 +115,6 @@ if( $x is Any( things... ) ) { ... }
 In any case, as junctions behave like values, they do not require special syntax like the block-invoking keywords proposed here, so they can be provided by regular function-call syntax from regular modules.
 
 ## Open Issues
-
-* There could be anything up to five new operators added by this idea. Do they all get their own named feature flags? Do they all live under one flag?
-
-* Should the flag be called `any`? That might be confusing as compared to the `:any` import tag which would request all features.
 
 ## Copyright
 


Grinnz:
Should it be called list_utils if it only includes any and all? I would say they are the most commonly used that require special syntax, but what if we want to add first or reduce in the future? Or would we consider whether to add them as part of this PPC?

Grinnz:
Well consider this my vote for adding 'first' at least. reduce I think needs a better way than $a and $b but if that can be resolved that would be nice to have too. none and notall are simple inversions of any and all so I'm not sure they're worth the feature bundle keywords.

Original PR for discussion: https://github.com/Perl/PPCs/pull/50

guest20:
There's a couple of fun options under "Deprecated and removed variables" in perlvar, I like `$*` best, though `$#` is pretty tempting too:

```perl
my $sum     = reduce { $* += $_ } @numbers;
my $grouped = reduce { $#->{ $_ } ++ } @numbers 
```

guest20:
> I don't think that giving a new meaning to a deprecated variable is the right approach.

We know they're already valid syntax and they've been available since 5.10 (perlvar)

Flavour-wise, I'd like it to be a single character variable so I don't have to repeat myself like I do in:

```perl
my $sum; $sum += $_ for @numbers; 
```

>  annoyance with reduce when the default for the accumulator is a hashref 

Can't one trust autoviv deal with the collector becoming a hash-ref?

book:
What about `$^a` and `$^b`? They have a the benefit of an obvious identification to the left and right side of an operation. And all the `$^` variables are already reserved.

leonerd:
If we want a `first` I can consider that, but lets do that in its own discussion.

I wanted to clarify here first what the feature/warning name would be, and also the alias/copy/readonly-copy behaviour of `$_` in the blocks. Can we discuss that, so I can finish and merge the work so far, before I consider the next bit? ;)

haarg:
Differing in behavior from grep/map would be extremely surprising. If map/grep use aliases, any/all should as well.

book:
> Differing in behavior from grep/map would be extremely surprising. If map/grep use aliases, any/all should as well.

My second-hand memory of this is that one is using aliases and not the other.

book:
OK, I think I'm wrong:

## `grep`

```
$ perl -E '@a=0..2; say for grep $_++, @a; say for "---", @a'
2
3
---
1
2
3
```

## `map`

```
$ perl -E '@a=0..2; say for map $_++, @a; say for "---", @a'
0
1
2
---
1
2
3
```
They both look like they're aliasing.

book:
> `$_` is an alias inside both grep and map. grep returns a list which contains aliases to the input list elements. map returns copies.

OK, that's what memory was about: the returned values.

If these are meant to be analogues of `map`/`grep`, they should probably have the same behaviour indeed.

leonerd:
My thinking here was that the aliasing behaviour is basically never used intentionally and any time it comes up it's always been a bug that has caused accidental modification. The fact that these new operators already don't support the `op EXPR, LIST` deferred expression syntax makes them feel a little different, so I thought that a second safety-oriented difference would be OK.

Alternatively, this could be something we fix later if/when we get around to adding named lexicals to the blocks. Perhaps a `:ro` attribute?

```perl
if(any my $x :ro { $x is now readonly copy of list } @things) { ... }
```

leonerd:
> Why not? Many people love to use the `EXPR` version of `map` and `grep`.

This was already mentioned in the original PPC0027.

Avoiding that syntax avoids a bunch of parser ambiguities that makes the-below easier to handle.

> Also, should these eventually support the proposed topicalization behaviour for `map` and `grep` ([PPC 0023](https://github.com/Perl/PPCs/blob/main/ppcs/ppc0023-map-grep-with-topic.md))?
> 
> Meaning:
> 
> ```
> op EXPR, LIST
> op BLOCK LIST
> op my VAR BLOCK LIST
> ```

Yes - the intention is that these can easily support that final idea. By not supporting the deferred expression form, those become easier to handle.

leonerd:
> †. ... though `first` seems like a sketchy name, since `last SAME_BLOCK, SAME_LIST` isn't going to give you the last match

`first` is already very-well established as the name for this:

https://metacpan.org/pod/List::Util#first

https://metacpan.org/pod/List::Keywords#first

leonerd:
As the idea for read-only copies of data in `$_` seemed unpopular, I've moved it into a suggestion for Future Scope; we can come back and think about that as another feature at another time.

What is the decision on feature flag name for these operators? Are we sticking with `any_all` for this, and accepting that `first` or `reduce` or whatever would each get their own flags? That does feel like a lot of new flags but perhaps that's OK?

book:
> What is the decision on feature flag name for these operators? Are we sticking with `any_all` for this, and accepting that `first` or `reduce` or whatever would each get their own flags? That does feel like a lot of new flags but perhaps that's OK?

We discussed this during the PSC meeting today. Assuming we're not constrained by the ever-growing number of feature flags, there should simply be one flag per symbol (so `any`, `all`, `first`, and `reduce`).  This will make it easier to add more symbols of this kind in the future.

We also looked at the [policy](https://perldoc.perl.org/perlpolicy), and we want to make it possible to progress quickly on this, so that the features can be marked as experimental (because of the change in behavior from the `List::Util` versions) in the next stable release, and taken out of experimental in the following one.

leonerd:
> We discussed this during the PSC meeting today. Assuming we're not constrained in the ever-growing number of feature flags, there should simply be one flag per symbol (so `any`, `all`, `first`, and `reduce`).

OK; I can do that. Though perhaps we should add some specific wording in the docs pointing out the case for `all`, that it is one specific symbol called "all", and not all of the features/warnings:

```perl
use feature 'all';
no warnings 'experimental::all';
```

book:
> Though perhaps we should add some specific wording in the docs pointing out the case for `all`, that it is one specific symbol called "all", and not all of the features/warnings:

This makes me realize that `all` is unfortunate as a feature name... Maybe that's a good argument for `any_all`, `first`, `reduce`...?

@haarg, @ap: what do you think?

leonerd:
My personal thought is that I think it's fine.

People really shouldn't be doing `use feature ':all'` anyway, because that's just a footgun waiting to happen. Yes it is a bit confusing that "all" doesn't mean all the features, it means one feature called "all", but I think that's less confusing overall for the consistency of the fact that it enables an operator called "all".

I've added a small extra detail to the docs in `feature.pm`:

```pod
B<Note:> remember that this enables one specific feature whose name is C<all>;
it does not enable all of the features.  This is not C<use feature ':all'>.
For that, see the section below.
```

guest20:
To me it feels like being defensive in some POD doesn't really address the usability issue of calling a single feature "all"

leonerd:
PR https://github.com/Perl/perl5/pull/22773 is currently paused awaiting a resolution on this question. At present it's implemented with two separate feature flags, named `all` and `any`. Despite the potential for confusion, I think this is the best choice as it nicely relates named feature flags with the names of the operators each enables.

Grinnz:
I don't particularly like the conflation with 'all' but at the same time, I've never seen the ':all' flag used and frankly it should be deprecated (in both the use and no case, given our negative features).

leonerd:
@Grinnz Indeed - I think especially in the case of `feature.pm`, it's very unlikely anyone would actually use `:all` the features, precisely because of those negative ones. I think the chance of a collision of semantics here is low enough not to worry.

ap:
> My thinking here was that the aliasing behaviour is basically never used intentionally and any time it comes up it’s always been a bug that has caused accidental modification.

I use it all the time. And I would absolutely expect it from list ops that are essentially specializations of `grep`. Sometimes the aliasing is undesirable and requires making a copy explicitly, which is slightly inconvenient, but an easy minor alteration of the code – whereas if aliasing was needed but copying was the available behavior, the only way out is switching to an entirely different approach (e.g. iterating over indexes instead of values).

PR #62, opened 2024-11-29
PPC0025: Update ID

johannessen:
Just a very tiny thing, but it kinda jumped out at me when first looking at the document.

### DIFF
--- a/ppcs/ppc0025-perl-version.md
+++ b/ppcs/ppc0025-perl-version.md
@@ -3,7 +3,7 @@
 ## Preamble
 
     Author:  Aristotle Pagaltzis <pagaltzis@gmx.de>
-    ID:      ARISTOTLE
+    ID:      0025
     Status:  Draft
 
 ## Abstract


book (approved):

guest20:
Maybe renaming files to include sequential numbers in their name, and contents and pr subjects is not a git / github friendly process?

haarg:
We have been discussing how we could improve the PPC process. Spreading conversations across PRs and the mailing list is problematic, and the handling of statuses and PPC# allocation isn't working well. No concrete conclusions yet though.

book:
> Maybe renaming files to include sequential numbers in their name, and contents and pr subjects is not a git / github friendly process?

The rename happens exactly once.

Assigning the PPC ID could be the last step of the initial MR, where we communicate to the author which filename to use before we merge.

guest20:
The procedure (from an uninvolved party):
The PR shouldn't⸸ be merged until a unique⸸, higher than the biggest of the other numbers⸸, by exactly one⸸ is chosen (by whom?) and added to the `ID: ` line in the file content in the PR⸸ and the file is renamed⸸. The number has to be unique⸸.

So, with all these chances for "oops", what's the win?

Referencing "PPC00025" gives the reader some text to search for and maybe find it‡... however:
- "PPC00025" will only find this comment - it has an extra zeros. 
- "PPC0025" does find the ppc, but only be because it's in the file name
- `ID: 0025` (from the template) doesn't match "PPC0025"⁑

...so you can only find a PPC "by id" after it's merged and renamed. This has [been inconvieniant](https://github.com/Perl/PPCs/issues/63#issuecomment-2509768692). 

The slug format is maybe a good shorthand for Change logs, or IRC, maybe sms/twitter where character counts are restricted, maybe, but even on a mailing lists it's better manners to link directly to the PR/issue/file using an http url

Whatever value all this extra book keeping⁂ is producing it can readily be replaced with "just the date"† and all the manual steps ✨ just go away ✨ 

IDK, maybe there's some machinery I'm not seeing, and this is actually useful, but it just feels like fighting the tools to me.


__ 
⸸. an unenforced, unchecked manual step
⁑. so these aren't even git grep'able
‡. As long as it's not part of a link, or pop up, or some other place in the UI that prevents a user selecting it. This is even worse on a touch-only devices.
†. It can be any date, since they automatically sort chronologically - even if it's the date of the first meetup you joked about it. 
⁂. heh. @book keeping.

ap:
> Just a very tiny thing, but it kinda jumped out at me when first looking at the document.

LOL I don’t know how that happened. Thanks for the catch.

PR #67, opened 2025-01-13
Add a simple website

davorg:
@book asked me to look at using GitHub Actions/Pages to put a simple website in front of this repo. This is what I've knocked up in the last couple of hours. It's not particularly pretty, but I figure it's a start and now we have the skeleton in place, we can iterate on it to make improvements.

If you decide to merge this, there's a bit of GitHub Pages admin needed. I'm happy to talk you through that or do it myself if you give me appropriate permissions on this repo.

You'll need to decide on a domain to host it (or you can use the GitHub default which would (I think) be https://perl.github.io/PPCs. In the meantime, my fork is deployed at ~https://ppc.perlhacks.com/~ https://davorg.dev/PPCs/ (moved it) for you to look at.


jkeenan:
The **Status** column should have definitions for the various statuses.  I wondered, "Does 'Implemented' mean that the PPC has appeared in a production release?  If so, which one (with date)?"

Otherwise, I appreciate the simplicity of the design.

davorg:
As I said, this is just a skeleton - more of a proof of concept. There are plenty of things that the PSC could add

davorg:
@guest20 The content is for the PSC to think about, My mission here was to demonstrate how it might work - which I think I've done.

haarg:
I think putting this at https://perl.github.io/PPCs will be fine.

guest20:
@davorg In that case I'm pleased to report that your PR has definitely convinced me that the internet is a suitable vehicle for templated hypertext documents.

Grinnz:
I think that this is a great idea for visibility into the process and the decisions being made, and I agree that the default github.io URL serves it well as a self-documenting indication of where the PPCs are maintained.

I would suggest moving the "Title" column to directly after the PPC number, to make it easier to scan for their names.

book:
> As I said, this is just a skeleton - more of a proof of concept. There are plenty of things that the PSC could add

And thank you very much for this.

I intend to spend some time playing with your proof of concept to organize the documents in a way that is more suitable to the process: the most important documents are the ones that are still being reviewed, or still being implemented. The PPCs for proposals that have either been rejected or added to Perl only serve as historical reference.

rwp0:
Thanks, two minor thoughts/suggestions while checking the design of the fork:

image

1) Unlink "as" in "as JSON" since "JSON" is enough already.

2) `@XXXX` hurt the eyes, remove the email domain part altogether.

Again, thank you very much for this is a worthy addition/integration IMHO.

esabol:
I think this is definitely needed, and it's a wonderful contribution.

Minor nitpicks of the web page itself:

1. Add a period after "as JSON".
2. Change "Github" to "GitHub".

davorg:
I've been quietly making some small improvements. A lot of the suggestions I've seen here have already been implemented. See the various commit titles for details.

And just now, I've added simple Datatable support, so the table on the front page can be paginated, sorted and filtered.

karenetheridge:
There are no timestamps.. I think the most relevant timestamp would probably be the date of last comment/commit/update, which surely is available in the github pull request metadata somewhere?

guest20:
@karenetheridge 

PR info isn't going to be a great choice here:
- You don't want to trigger a build every time somebody comments, so you'll end up with last-comment dates from the most recent merge
- PRs/Comments aren't in git, so either the site, or the build need to query github's graphql api

The question is "what are you looking to rank up?"... maybe sorting on Status first (to put the most sensational new ideas at the top), then on commit date (to keep the newest top'est) would make the most sense?

guest20:
"0006 	Module Loading with "load\_module""

That `\` in there makes me feel like there's some kind of (possibly unsafe) markdown shenanigans going on, obviously the `_` needs to be escaped in markdown so it's not turned into an `<em>`, but the list on the home page is supposed to be rendered markdown, not markdown source. 

(I just fixed an unrelated bug in unrelated code where my table-of-contents logic contained literal `[](http://...)` because it forgot to parse headings for links before putting them in the ToC so I might be being paranoid here)

ttlib/page.tt
```diff
@@ -0,0 +1,105 @@
...
+<html lang="en" class="h-100">
+  <head>
+[% IF base -%]
+    <base href="[% base %]">
+[% END -%]
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="description" content="">
+    <meta name="author" content="Mark Otto, Jacob Thornton, and Bootstrap contributors">
+    <meta name="generator" content="Hugo 0.84.0">
```
JRaspass, line 11: Cargo culted? We're not using Hugo here.

guest20: Maybe it's a security measure to distract vulnerability scanners?

davorg: Yeah, I'll fix that. It's there because I copied the code from a Bootstrap example site.

I really wouldn't describe it as cargo-culting - I know exactly what all of those lines do 😊

lib/PPC.pm
```diff
@@ -0,0 +1,86 @@
+use experimental qw[builtin class signatures];
```
JRaspass, line 1: Maybe add a `use v5.38` at the top, like the build script, I don't think anything in this file currently turns on strictures/warnings. As a benefit the the 5.38 bundle has `module_true` so you can drop the `1;` from the end of this file.

davorg: Good suggestion. Thanks. Done.

davorg: Also fixed the warning that this revealed :-/

JRaspass: Nice, while we're here is there any reason not to use the latest Perl, 5.40? Looks like the GH Action uses "latest".

Using 5.40 would present a few opportunities to remove boilerplate:

Builtin is not longer experimental, so we can just do:
```perl
use experimental qw[class signatures];
```
Builtin has version bundles too now which are automatically imported by the corresponding use version line so we can drop this line completely:
```perl
use builtin 'trim';
```
And the biggest win is we can apply `:reader` to our fields and drop our handed rolled readers:
```perl
field $author  :param :reader = '';
field $id      :param :reader;
field $slug    :param :reader;
field $sponsor :param :reader = '';
field $status  :param :reader;
field $title   :param :reader;
```
Remove:
```perl
method author  { return $author }
method id      { return $id }
method slug    { return $slug }
method sponsor { return $sponsor }
method status  { return $status }
method title  { return $title }
```

davorg: > is there any reason not to use the latest Perl, 5.40?

Honestly, just laziness. The system I was working on had 5.38 installed :-)

But, yeah, I should update that.

davorg: Done

davorg:
@Perl/perl-steering-council Is there anything else you need before merging this?

book:
Thanks, I rebased your branch, and applied it as 1d5e7a44c713d52c2b089ee4194a67ee8806c969.

esabol:
> Thanks, I rebased your branch, and applied it as [1d5e7a4](https://github.com/Perl/PPCs/commit/1d5e7a44c713d52c2b089ee4194a67ee8806c969).

The URL (https://perl.github.io/PPCs) for this website probably needs to be added to https://github.com/Perl/PPCs/blob/main/README.md. I also recommend putting it in the "About" box of the project on GitHub.

PR #68, opened 2025-01-22
PPC0033: method-like calls to lexical subs

haarg:
Add a syntax `$object->&method(@args)` that calls `method` as a sub, to allow a nicer syntax for private methods.

### DIFF
--- a/ppcs/ppc0033-ampersand-method-calls.md
+++ b/ppcs/ppc0033-ampersand-method-calls.md
@@ -1,0 +1,129 @@
+# Allow calling subs using method-like syntax
+
+## Preamble
+
+    Author:  Graham Knop <haarg@haarg.org>
+    ID:      0033
+    Status:  Draft
+
+## Abstract
+
+Add a syntax for calling lexical methods that is similar to the existing
+method syntax, rather than needing to call them like subs. The new syntax
+`$object->&method(@args)` would be equivalent to sub call to `method`, but
+using a syntax closer to method calls.
+
+## Motivation
+
+When using a lexical sub inside an object class, the subs must be called as
+`method($object, @args)`. This works, but for many people, this feels wrong
+as a way to call something thought of as a private method. This has resulted
+in some people using syntax like `$object->${\&method}(@args)`, or continuing
+to use code refs stored in a lexical variable.
+
+## Rationale
+
+`$object->method(@args)` always does a method lookup, ignoring the local
+context. Changing this would result in a lot of broken code. Changing this
+only for new code (via feature or other lexical effect) would introduce
+confusion and would also require some new (or pessimised) syntax to allow
+the previous behavior.
+
+`$object->&method(@args)` would be an unambiguous way to call the `method` in
+the current scope, whether a lexical or package sub.
+
+A truly private method does not need to do any lookup via `@ISA`, so being
+equivalent to a sub call makes sense.
+
+This also would pair well with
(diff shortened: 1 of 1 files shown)


ppcs/ppc0033-ampersand-method-calls.md
```diff
@@ -0,0 +1,129 @@
...
+my $foo = Foo->new;
+$foo->public;
+
+# this is an error, as there is no sub named private in scope
+$foo->&private;
+```
+
+## Prototype Implementation
+
+None currently.
```
leonerd, line 105: There's my entire CPAN module which experiments with this very syntax idea (though it does so in a kinda hacky way):

https://metacpan.org/pod/Object::Pad::LexicalMethods

ppcs/ppc0033-ampersand-method-calls.md
```diff
@@ -0,0 +1,129 @@
...
+`$object->&method(@args)` would be equivalent to sub call to `method`, but
+using a syntax closer to method calls.
+
+## Motivation
+
+When using a lexical sub inside an object class, the subs must be called as
+`method($object, @args)`. This works, but for many people, this feels wrong
+as a way to call something thought of as a private method. This has resulted
+in some people using syntax like `$object->${\&method}(@args)`, or continuing
+to use code refs stored in a lexical variable.
```
karenetheridge, line 22:
```suggestion
to use code refs stored in a lexical variable (`$object->$method(@args)`).
```

ppcs/ppc0033-ampersand-method-calls.md
```diff
@@ -0,0 +1,129 @@
...
+
+    sub public ($self) {
+
+        # call via lexical method is allowed
+        say "my field: " . $self->&private;
+        # exactly equivalent to:
+        say "my field: " . private($self);
+
+
+        # can also be used to call package methods. But won't lookup via @ISA
```
karenetheridge, line 89:
```suggestion
        # can also be used to call package methods. But won't look up via @ISA
```

haarg:
I will clarify that this should work for public methods in the class syntax, but I had put that in the future scope section because `my method` is not yet supported. This PPC would be a way to make lexical methods in the class syntax viable, but support for lexical methods is a separate issue from the call syntax.

leonerd:
I've had a go at implementing the `my method` syntax in https://github.com/Perl/perl5/pull/22940 as it was an oversight not to have done it anyway. That said, we might want to wait until implementing this PPC as well to get them both in at once, to avoid an intermediate situation whereby `my method` exists but `->&` does not, leading to having to suggest the "just call it like a sub" workaround.

ppcs/ppc0033-ampersand-method-calls.md
haarg: Added a link to this section.

ppcs/ppc0033-ampersand-method-calls.md
haarg: Added.

ppcs/ppc0033-ampersand-method-calls.md
haarg: Changed.

leonerd:
I'm definitely in favour of this. In fact if it were merged soon I'd like to get an implementation of it added in with my `my method` PR above. It makes sense to add them both together; as either without the other is rather pointless. So the PR is largely waiting on this idea.

PR #69, opened 2025-02-06
update the PPC status to be more consistent

book:
The different values for PPC status are somewhat inconsistent between the various documents in the docs/ directory, and the actual PPC documents.

For example, "Accepted" in the process document means "with the code merged as a new (probably experimental!) feature of perl", while in the current PPCs documents, it was used to mean "Accepted Proposal".

During PSC #178 2025-02-06, Aristotle and BooK discussed the issue, and proposed some new status names, with a goal of improving clarity.

This commit makes the following status renames:

    Draft       => Proposed
    Exploratory => Proposed
    Accepted    => Wanted
    Implemented => Incorporated

The discussion also revolved about the possible parallel existence of an idea (the PPC proposal) and an implementation. An implementation can exist while the PPC is still being discussed an evaluated, and help steer it. An implementation can be rejected, while the idea is still accepted.

### DIFF
--- a/ppcs/ppc0001-n-at-a-time-for.md
+++ b/ppcs/ppc0001-n-at-a-time-for.md
@@ -5,7 +5,7 @@
     Author:
     Sponsor: Nicholas Clark <nick@ccl4.org>
     ID:      0001
-    Status:  Implemented
+    Status:  Incorporated
 
 ## Abstract
 
--- a/ppcs/ppc0004-defer-block.md
+++ b/ppcs/ppc0004-defer-block.md
@@ -5,7 +5,7 @@
     Author:  Paul Evans <PEVANS>
     Sponsor:
     ID:      0004
-    Status:  Implemented
+    Status:  Incorporated
 
 ## Abstract
 
--- a/ppcs/ppc0006-load-module.md
+++ b/ppcs/ppc0006-load-module.md
@@ -5,7 +5,7 @@
     Author:  Ricardo Signes <rjbs@semiotic.systems>
     Sponsor: Ricardo Signes <rjbs@semiotic.systems>
     ID:      0006
-    Status:  Implemented
+    Status:  Incorporated
 
 ## Abstract
 
--- a/ppcs/ppc0007-source-encoding.md
+++ b/ppcs/ppc0007-source-encoding.md
@@ -5,7 +5,7 @@
     Author:  Ricardo Signes <rjbs@semiotic.systems>
     Sponsor: Ricardo Signes <rjbs@semiotic.systems>
     ID:      0007
-    Status:  Draft
+    Status:  Proposed
 
 ## Abstract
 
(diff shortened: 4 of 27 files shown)


haarg:
Fixing the statuses seems reasonable, but this seems like it's trying to establish a new process which isn't documented.

book:
> Fixing the statuses seems reasonable, but this seems like it's trying to establish a new process which isn't documented.

These statuses do not change the process, they try to clarify the state of each PPC by using clearer wording. I *think* we can use those and not betray the process that led the current PPCs where they are.

My intention was also to quickly merge the currently open PRs, and move them all to status "Proposed". I find that the current use of "Exploratory" in some PPCs does not match what's currently described in the current process.

Quickly merging PR for new PPCs would avoid lengthy discussions in the PR itself. We want the rejected PPCs to exist in the repository too, as an historical record of what was discussed in the past. Therefore they might as well be merged ASAP. @ap would rather like that the discussion happens on the mailing-list, but we know that use of GitHub is inevitable. It's possible that issues are a better medium for discussion than pull requests (which would cover copy editing and updates following the discussions).

I want to use wording for the statuses that makes it very clear what the current state of the PPC is, especially for the early and late stages:
* **Proposed**: the document has been proposed, and is being discussed by the community and the PSC.
* **Wanted**: consensus has been reached, the PSC likes the idea, and if someone proposed a reasonable implementation, it would be very likely accepted.
* **Incorporated**: an implementation of the PPC is now part of Perl.
 
You are right, though, that some of the discussion on Thursday revolved around maybe improving the process (and the format of the documents). @ap and I haven't reached consensus, and any significant change would go through a PR anyway.

Some of the discussion revolved around making a difference between the idea and the implementation. We were discussing some cases that not currently covered explicitly:
* the idea is accompanied with an implementation
* the possibility of rejecting an implementation, while still approving the idea

The next significant step would be to update the existing process document with those new status names; I expect the scrutiny with which I'll read the document will lead to a better understanding of the current process and some ideas on how to improve it.

Grinnz:
If more "quickly merging" of PPCs is to be done, I would suggest finding an accepted mechanism for discussing active PPC proposals as currently the PR is the only place housing such discussion.

book:
> If more "quickly merging" of PPCs is to be done, I would suggest finding an accepted mechanism for discussing active PPC proposals as currently the PR is the only place housing such discussion.

Once pull requests are out (except for small discussions about details of a change to a PPC), there are not that many possible places available for having a discussion. I think it comes down to:
* the mailing-list (we could trigger the discussion by announcing the proposed PPC has been merged)
* GitHub issues

I would also like to add a "Reference" section to all PPCs, with links to the various records of the discussion. That would make things easier for future code archaeologists.

Grinnz:
> > If more "quickly merging" of PPCs is to be done, I would suggest finding an accepted mechanism for discussing active PPC proposals as currently the PR is the only place housing such discussion.
> 
> Once pull requests are out (except for small discussions about details of a change to a PPC), there are not that many possible places available for having a discussion. I think it comes down to:
> 
> * the mailing-list (we could trigger the discussion by announcing the proposed PPC has been merged)

This sounds reasonable to me and would be most inclusive of stakeholders

PR #89, opened 2026-05-13
[PPC0035] Add document about Magic v2

leonerd:
Also updates some wording in PPC0029 for the new name "Magic v2"


leonerd:
As this hasn't received any objections (or in fact any comments at all) in the past nearly two months, I will merge it now.

PR #90, opened 2026-06-05
[PPC0036] Add document about Viral Value Magic

leonerd:


ppcs/ppc0036-viral-value-magic.md
```diff
@@ -0,0 +1,163 @@
...
+
+As this magic shape is strongly associated with the value stored in the SV, rather than the SV itself, it has quite different rules for copying, clearing, and performing other calculations than other magic shapes have.
+
+* When copying a value, this magic is propagated from the source to destination SV on calls such as `sv_setsv()` by invoking `infect`.
+
+* When clearing a variable, this magic is removed from the variable SV on calls such as `sv_undef()`. A `free` trigger function on the base magic may be invoked if present.
+
+* When calculating a new value as a result of one *or more* input values that have these annotations attached, the `infect` function on each annotation of each such value is invoked to copy it into the resulting destination. It may be that the result ends up with more than one annotation attached to it.
+
+* When a container variable is `local`ised, magic annotations of this shape are not copied into the new temporary container.
```
rabbiveesh, line 54: this behavior doesn't match with what i'd have in my head - i'd imagine that unless mutated, a `local`-ised copy would continue being the same thing.
perhaps our mental models are not aligned - could you spell out specifically why you want this behavior?

ppcs/ppc0036-viral-value-magic.md
```diff
@@ -0,0 +1,165 @@
...
+
+```c
+enum MagicShape {
+    ...,
+    MGv2s_SCALARVALUE,
+};
+
+struct ScalarValueMagicFunctions {
+    ..., /* existing fields, where shape == MGv2s_SCALARVALUE */
+    void (*infect)(pTHX_ SV *ssv, MAGIC *smg, SV *dsv, MAGIC *dmg);
```
Leont: I'm sure we can find a better name than that.

ppcs/ppc0036-viral-value-magic.md
Leont: We may need to have a sharper distinction between value and container magic before trying to make value magic viral.

ppcs/ppc0036-viral-value-magic.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Security Implications
+
+Normally, this section is used to explain around any potential security *problems* that might be introduced by the changes proposed by such a document. As is typical when considering adding new low-level abilities that modules can build on top of, the answer is usually "none foreseen, they would have to be considered by the modules being built".
+
+But a much better answer here is to observe that this proposed mechanism adds a great ability to introduce new *solutions* to security problems. As perhaps was already hinted at in the introduction, the mechanism proposed here acts similar to but is more flexible and powerful than perl's existing "taint" behaviour, which was primarily intended for its security-related purposes. In practice, that mode is seldom used, because of its lack of flexibility. Taint mode just stores a single boolean annotation about any particular value ("is this considered tainted?"), which affects certain entirely-internal functions and operations. In contrast, this proposal would easily permit the creation of a more flexible set of policies; see the "Supertaint" example below.
```
Leont, line 68: I think it's very different from taint, I'm not sure it can actually replace that logic; taint works much more indirect that this would.

ppcs/ppc0036-viral-value-magic.md
leonerd: When you `local` a variable, the internals of how it works means that you get a new SV that gets moved into place for where that variable was visible. This means that magics that ought to be considered part of the "container" (things like the special behaviours of the various dollar-punctuation variables) needs to be copied onto that new SV as well, so that "the SV visible as that variable" retains that behaviour.

Whereas, any behaviours that are part of the value currently stored in the SV should not get copied into that new one; they belong with the value rather than the variable that contains it.

ppcs/ppc0036-viral-value-magic.md
leonerd: Likely so; this was just a unique name that came to mind that would be pretty easy to search-replace later on.

ppcs/ppc0036-viral-value-magic.md
leonerd: > We may need to have a sharper distinction between value and container magic before trying to make value magic viral.

Perhaps so. I try to make more of a distinction of that in Magic-v2 (see the corresponding PPC0035). With v1 you basically just distinguish by being `PERL_MAGIC_ext` vs `PERL_MAGIC_extvalue` and that has effects on `local`isation, whereas magic-v2 has a specific `MGv2f_CONTAINER` flag to properly say "this relates to the container, not the value stored in it".

ppcs/ppc0036-viral-value-magic.md
leonerd: Maybe `propagate` is better? It's a slight overlap with the thing that exception objects do on rethrow...

cjac:
Strong support for this proposal!

I am currently working on building modules to elevate Perl as a first-class participant in the Google Cloud Platform (GCP) space. For modern enterprise cloud customers, automatic Data Lineage and observability are critical requirements, driven centrally by GCP Dataplex.

The "Viral Value Magic" concept is exactly what Perl needs to stay competitive in this space. Currently, tracking data moving through a Perl script (e.g., pulling from BigQuery, manipulating it, and uploading to Cloud Storage) requires verbose, intrusive instrumentation.

Having this automatic propagation built directly into the core interpreter would allow Perl to natively produce high-fidelity OpenLineage traces (including column-level lineage) effortlessly. These traces could be pushed directly to the Dataplex Data Lineage API, making Perl significantly more viable for modern, cloud-native data pipelines alongside peers like Python or Java. Since the implementation is essentially ready, I strongly urge the inclusion of this PPC.

ppcs/ppc0036-viral-value-magic.md
mohawk2: I think here an example would paint a thousand words.

leonerd: Ah so in actual fact it turns out that literally today, I reworked the localisation stuff anyway. Rather than the "container" flag, I got rid of that in favour of having a `localise` trigger function on the (scalar|array|hash) variable magic tables. By default it does nothing, so the magic disappears on localisation (i.e. matches the `PERL_MAGIC_extvalue` behaviour), but if present that is invoked so you can do something different. A handy function is also provided to put in there directly if you just want to copy the magic, so if you set that it's basically like having `PERL_MAGIC_ext` behaviour.

Zhtwn:
@cjac Ayup, value magic will be very useful for lineage tracking.

I'm working on CPAN modules that provide APIs that wrap the in-core value magic:
- [Scalar::ValueTags](https://metacpan.org/pod/Scalar::ValueTags): allows multiple clients to independently use value magic, and provides selectable propagation behavior and tag data structures
- [Scalar::ValueTags::LineageTracking](https://metacpan.org/pod/Scalar::ValueTags::LineageTracking): registers as client of `Scalar::ValueTags`, and propagates all of the unique tags from input variables onto output variables. It also allows the client to select whether tags use either an arbitrary unblessed ref or a serialized string. This allows propagation-heavy use cases to select the faster string-hash propagation, while propagation-light use cases can use somewhat-slower but more flexible arbitrary data structures.

I'd definitely appreciate any input.

nrdvana:
For the sake of completeness, I want to summarize here the overlap with Perl's overloaded objects, which I've now discussed both on p5p and IRC.

This proposal is really about attaching metadata to a value, which then gets copied alongside the value during assignments from one variable to the next, and sometimes across operations that compute a new value.  Perl already has a roundabout method of attaching metadata to a value by declaring an object (containing as much metadata as you like) and then adding overload magic to make the object reference behave like an immutable value.  This has been demonstrated in JSON::PP which uses boolean objects to represent true/false for round trips.

The overloaded-immutable-object mechanism gives almost entirely equivalent user-side behavior as value magic, but is slow, has lots of edge cases, and can't be composed of dissimilar metadata from multiple authors.  For instance, calling `ref` on a JSON::PP::Boolean returns a true value which may confuse code that was expecting only a plain scalar.  Also, every time you access the value of the overloaded object it calls a function, which is significant overhead.  It is also challenging to write the complete collection of overload methods so that the metadata can be propagated to new value, and creating the new value requires constructing an object, which is yet more overhead.  Overloaded-immutable-objects are likewise capable of implementing the lineage-tracking use cases described, though would perform so badly that likely nobody would want to.

I think that value-magic as proposed offers much better performing mechanism for attaching metadata to values, so it would be worthwhile to add.

One limitation I see is that value magic's callback won't know what operation was just performed on the input values to create the new value.  "viral metadata" doesn't care, but if you were trying to do something more nuanced like attach units to numbers, you would need to know what operation had just been performed.

```
my $x= 10.500;
set_value_metadata($x, SI_UNIT, "meters");
set_value_metadata($x, SIGNIFICANT_DIGITS, 5);
my $y= 4.2;
set_value_metadata($x, SI_UNIT, "seconds");
set_value_metadata($x, SIGNIFICANT_DIGITS, 2);
$x *= $y;
```

In that example, the callbacks for SI_UNIT would not know whether the new value should be meters/second, meter-seconds, or warn that two dissimilar units had been added.  Likewise, the SIGNIFICANT_DIGITS needs to know whether it was basic arithmetic or an operation like pow(), or so on.

```
my $s= "\xC4\x80";  # encoding of char 0x100
set_value_metadata($s, CHARSET, "UTF-8");
my $l= "\xA9";      # copyright symbol in latin1
set_value_metadata($s, CHARSET, "LATIN-1");
my $u= "\x{100}";   # actual char 0x100
set_value_metadata($u, CHARSET, "UNICODE");
$u .= $s . $l;
```

In this example, the value metadata can take a pretty good guess that mismatched character encodings have been combined together into the same string, and warn about it.  That's an improvement over the status quo, but still doesn't enable code to "do what I mean".  This might be asking too much from a value metadata feature, but might be nice if value metadata had a chance to do something like overload does without needing to convert all of these strings to immutable objects.

One of the comparisons made in the proposal was the taint feature.  The taint feature would likewise not be implementable with value magic as defined so far, because taint needs to turn off when certain operations are applied, like testing with a regex or extracting a capture from a regex.  The callback would need to know which operation happened.

This raises other questions, like whether value metadata infects from a base string to its regex captures, or whether it infects to every result of split(), whether it passes through the sqrt() function, whether authors have any control over which operations are infectious, and whether that list could be expanded to include user-defined functions that the value metadata author doesn't have control over.

Another question is whether the overloaded-immutable-object pattern could itself be improved with a better, more efficient, and more extensible back-end.

leonerd:
@nrdvana 

> One limitation I see is that value magic's callback won't know what operation was just performed on the input values to create the new value. "viral metadata" doesn't care, but if you were trying to do something more nuanced like attach units to numbers, you would need to know what operation had just been performed.
> 
> ```
> my $x= 10.500;
> set_value_metadata($x, SI_UNIT, "meters");
> set_value_metadata($x, SIGNIFICANT_DIGITS, 5);
> my $y= 4.2;
> set_value_metadata($x, SI_UNIT, "seconds");
> set_value_metadata($x, SIGNIFICANT_DIGITS, 2);
> $x *= $y;
> ```
> 
> In that example, the callbacks for SI_UNIT would not know whether the new value should be meters/second, meter-seconds, or warn that two dissimilar units had been added. Likewise, the SIGNIFICANT_DIGITS needs to know whether it was basic arithmetic or an operation like pow(), or so on.

Hmmm. Yes; I see where you're aiming at there. Ideally it would be nice for those to be aware of what's going on.

As it stands, the design started off from a very simple extension of the vstring magic in `sv_setsv`, which just copies from one source to one destination. That worked fine for simple copies around the place, but it doesn't handle calculating derived values.

The way we currently do derived values is a very simple "throw everything into a bag" approach: Namely, every pp function that implements an op simply grabs the total set of all annotations on all of its input arguments, then throws all of those back out onto all of its output results. There's a bunch more special-case logic on top to handle weird things like regexps, but overall it's a simple scheme that works regardless of the shape of the op - whether it has 1, 2, or many inputs, and how many outputs it produces.

It is tempting to consider more powerful mechanisms that would be aware of the "shape" of the ops being involved. Perhaps you'd have a whole set of "propagate" trigger functions for different op classes:

```c
void (*propagate_unop)(..., SV *dsv, SV *ssv, OPCODE op);            /* for $dsv = OP $ssv */
void (*propagate_binop)(..., SV *dsv, SV *lsv, SV *rsv, OPCODE op);  /* for $dsv = $lsv OP $rsv */
```

but then almost immediately you hit upon a limitation of the C-shaped API in how you'd handle listops like `OP_JOIN`.

Then you start to get questions about how to handle ops that return multiple values - `OP_SPLIT` being a classic.

OK so now maybe you'd consider maybe having just two functions, for "incoming args" vs "outgoing results", and tell each one the opcode and maybe some sort of argument index:

```c
void (*propagate_arg)(..., SV *dsv, SV *ssv, OPCODE op, size_t argix);
void (*propagate_ret)(..., SV *dsv, SV *ssv, OPCODE op, size_t retix);
```

The intention here is that you'd invoke `propagate_arg` on each incoming argument with increasing values in argix, and then `propagate_ret` on each outgoing result with similar.

But is that too much of a complication to be able to use in a sane way? It would be a fun thought experiment to see how you might actually implement the units-tracking metadata behaviour around this kind of thing.

I think it's possible it could work, but it would be quite a more complex shape than the one proposed ((and implemented - see below)) here; something quite beyond what is necessary for use-cases like lineage tracking.

> ```
> my $s= "\xC4\x80";  # encoding of char 0x100
> set_value_metadata($s, CHARSET, "UTF-8");
> my $l= "\xA9";      # copyright symbol in latin1
> set_value_metadata($s, CHARSET, "LATIN-1");
> my $u= "\x{100}";   # actual char 0x100
> set_value_metadata($u, CHARSET, "UNICODE");
> $u .= $s . $l;
> ```
> 
> In this example, the value metadata can take a pretty good guess that mismatched character encodings have been combined together into the same string, and warn about it. That's an improvement over the status quo, but still doesn't enable code to "do what I mean". This might be asking too much from a value metadata feature, but might be nice if value metadata had a chance to do something like overload does without needing to convert all of these strings to immutable objects.

It was a fairly firm design decision that "value magic" wouldn't be able to alter the values that it is applied to. It is purely a metadata layer that sits on top of the perl-visible values; but any operation you'd perform on those, you could just not look at the value magic on the result and it would look the same.

I think if you did want to consider some sort of fancy automatic encoding annotations on strings which would actually alter the values stored in those strings, that would be quite special-purpose enough on strings in particular to need to create a new magic-v2 shape to handle that with.

> This raises other questions, like whether value metadata infects from a base string to its regex captures, or whether it infects to every result of split(), whether it passes through the sqrt() function, whether authors have any control over which operations are infectious, and whether that list could be expanded to include user-defined functions that the value metadata author doesn't have control over.

As a specification, it doesn't really go into detail beyond saying it should generally copy around in all the ways you'd expect. But implementation-wise (see https://github.com/Perl/perl5/pull/24508) it very-much does make sure to copy through regexp captures, `split()`, `sqrt()` and allsorts else. In fact it was a rather nontrivial part of the entire implementation to ensure it copies properly though regexp capture buffers. :)

nrdvana:
> OK so now maybe you'd consider maybe having just two functions, for "incoming args" vs "outgoing results", and tell each one the opcode and maybe some sort of argument index:
> 
> ```c
> void (*propagate_arg)(..., SV *dsv, SV *ssv, OPCODE op, size_t argix);
> void (*propagate_ret)(..., SV *dsv, SV *ssv, OPCODE op, size_t retix);
> ```
> 
> The intention here is that you'd invoke `propagate_arg` on each incoming argument with increasing values in argix, and then `propagate_ret` on each outgoing result with similar.
> 
> But is that too much of a complication to be able to use in a sane way? It would be a fun thought experiment to see how you might actually implement the units-tracking metadata behaviour around this kind of thing.
> 
> I think it's possible it could work, but it would be quite a more complex shape than the one proposed

What if you just add a pointer to the opcode right from the opcode tree, and a pointer to the stack?  It would be fragile across versions, but wouldn't require any extra work to set up the callback, and authors could do as much or as little inspection as they needed.

I assume you're calling the existing infect() function for each instance of value magic on each input argument and doing that once for each output argument?

For the SI_UNIT example, it could then be (pseudocode-ish)

```
infect {
  /* combining magic? */
  if (get_si_unit_magic(dst)) {
    if (op is division) {
      set_si_unit_magic(dst, build_fraction(get_si_unit_magic(dst), get_si_unit_magic(src));
    }
    else if (op is multiplication) {
      set_si_unit_magic(dst, build_product(get_si_unit_magic(dst), get_si_unit_magic(src));
    }
    else if (get_si_unit_magic(dst) != get_si_unit_magic(src)) {
      warn("combining dissimilar units %s with %s", get_si_unit_magic(dst), get_si_unit_magic(src));
      set_si_unit_magic(dst, SI_UNIT_INDETERMINATE);
    }
  }
  else {
    /* don't care what opcode if dst doesn't have magic yet */
    set_si_unit_magic(dst, get_si_unit_magic(src));
  }
}
```

cjac:
I'm trying to find tools that perform the functionality of SAST scanners, and that appears to be another area where this PR would be very helpful.

---

# PR Review: PPC0036 - Viral Value Magic

## Overview & Strategic Value

This proposal for **"Viral Value Magic"** (Propagating Metadata) addresses a critical gap in modern Perl capabilities, particularly for enterprise data pipelines and cloud observability (e.g., GCP Dataplex, OpenLineage). The ability to propagate metadata transparently alongside values provides a generalized, highly efficient mechanism that serves two major industry demands: **Data Lineage** and **Advanced Application Security (Taint Protection)**.

---

## Technical Insights & Synergy

### 1. Duality of Lineage and Security (Taint Tracking)
The underlying concept of Viral Value Magic is mathematically and conceptually isomorphic to **Taint Tracking**:
*   **Data Lineage Use Case:** Attaching Tags (e.g., `Source: BigQuery.Users.Email`) to a value. As it is concatenated or transformed, the tag travels with it. Upon hitting a destination (e.g., Google Cloud Storage), the tag is read to register the pipeline path automatically.
*   **Security (Taint) Use Case:** Attaching Tags (e.g., `Input: Untrusted_User_Input`) to a value. If this value reaches a critical sink (e.g., `sysopen` or a Shell execution handler) without being laundered, the runtime can reject execution.

Integrating this as/around "Value Tags" (as discussed with CPAN contributors) effectively equips Perl with a **built-in Runtime Dataflow Engine**.

### 2. Filling Gaps Left by Static Analysis (e.g., `Perl::Critic`)
Static analyzers (like `Perl::Critic`) are excellent at catching syntactic anti-patterns but suffer from "context blindness"—they cannot easily trace runtime data execution graphs across deep function stacks. 
Viral Value Magic fulfills this by providing **Runtime Verification**. This allows maintaining strict policy assurances that cannot be caught statically but are paramount in highly regulated environments.

### 3. Implications for Modern Toolchains (LLMs & Generated Code)
For environments utilizing LLMs or algorithmic generation to build driver wrappers and SDKs, this feature provides a safety net against boilerplate leaks. Modern engines can look at these tags and auto-append lineage boundaries or apply fallback defenses transparently, reducing the cognitive overhead on standard developers to "remember" calling logging instrumentation modules.

---

## Core Focus Areas for Merge

As development progresses, the primary focus should iterate on keeping the "Viral" aspects precise:
*   Ensure metadata retention holds rigorously across **Regular Expression Captures (captures, `split()`, etc.)**, **string concatenation** boundaries, and **custom math** derivatives.
*   Keeping overhead low for standard operations (like standard copies without magic) remains essential to avoid regression against standard C algorithms.

This mechanism adds deep dimension tracking capability to Perl value flows, which will solidify its argument as a first-class citizen in dynamic data engineering workflows. Strongly recommended to advance execution along this lineage tracking framework.

ppcs/ppc0036-viral-value-magic.md
cjac: here are some verbs:

* impart
* derive or derive_from
* carry or carry_forward
* transfer
* flow or flow_to
* cascade
* stream
* percolate
* tag or tag_with
* annotate
* imprint
* diffuse
* disseminate
* transfuse

cjac:
Can we take the SE Linux label or Windows equivalent or FreeBSB equivalent or Mac OS equivalent and put it into some metadata thing associated with a variable? Sorry a an SV I think is what it is scalar value 

Would it be possible to take that information and place it into some sort of field? Maybe it would be better if it was a more special purpose field than just the general magic metadata field. Maybe there should be some sort of index which uses the magic Candida Fields 

Sorry I don't know the code. All that well yet. I'm kind of spitballing and thinking about how it might fit into my use cases 

Hopefully this makes sense

